PR #6, opened 2021-12-22
WIP: Adds Merkle cache for `Vector` type

ralexstokes:


ralexstokes:
closing as it is stale, but the general direction is worth iterating on

PR #10, opened 2022-02-14
Update bitvec to 1.0.0

jmcook1186:
PR updates BitVec version to 1.0.0 and fixes bugs arising from this change.

All tests pass.

### DIFF
--- a/ssz_rs/src/merkleization/cache.rs
+++ b/ssz_rs/src/merkleization/cache.rs
@@ -1,5 +1,5 @@
 use crate::merkleization::Node;
-use bitvec::prelude::{bitvec, BitVec};
+use bitvec::prelude::{bitvec, BitVec, Lsb0};
 
 #[derive(Default, Debug, Clone)]
 pub struct Cache {
@@ -12,7 +12,7 @@ impl Cache {
     pub fn with_leaves(leaf_count: usize) -> Self {
         Self {
             leaf_count,
-            dirty_leaves: bitvec![1; leaf_count],
+            dirty_leaves: bitvec![1,],
             ..Default::default()
         }
     }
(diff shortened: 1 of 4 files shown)


ssz_rs/src/merkleization/cache.rs
```diff
@@ -12,7 +12,7 @@ impl Cache {
     pub fn with_leaves(leaf_count: usize) -> Self {
         Self {
             leaf_count,
-            dirty_leaves: bitvec![1; leaf_count],
+            dirty_leaves: bitvec![1,],
```
ralexstokes, line 15: does this make a bitvector of length `leaf_count` that has all bits set?

ralexstokes (changes requested):
nice work!

from scanning the commits, it looks like we only need d91c0e3f087cfb2c2ddca011693612e191e3df26 in this PR

can you rebase this branch to drop the other 3?

ssz_rs/src/merkleization/cache.rs
jmcook1186: Now updated (055129e) to explicitly define bitvec length to be equal to leaf_count.

jmcook1186:
Rebased to remove uneccessary commits. 

New commit revises definition of `dirty_leaves`

BitVec 0.22 -> BitVec 1.0.0 seems to include a breaking change for how `bitvec!` macro works in the Cache definition (`/merkleize/cache.rs` line 13).

The original way now throws an error because the macro expects a constant and we need to pass in a variable (`leaf_count`).

I couldn't find a more elegant solution to this other than moving the bitvec construction to precede the struct definition.

I also included two `assert_eq's` to demonstrate the length and content are OK. These can be removed later.

PR #25, opened 2022-11-17
add `no-std`

claravanstaden:
- Adds `no-std` support by adding `std` module that imports from `std` or `core`/`alloc` depending if std feature is toggled.
- Removes `thiserror` crate, no support for no-std yet, but in the works. It depends on `Error` being included in the core crate: https://github.com/rust-lang/project-error-handling/issues/3 and `thiserror` issue: https://github.com/dtolnay/thiserror/issues/196


ralexstokes:
it looks like some of the changes in here were done in #29; please rebase `main` when you get a chance -- very excited to dig into this PR :)

ssz-rs/src/uint.rs
```diff
@@ -93,6 +92,7 @@ impl U256 {
     }
 }
 
+#[cfg(feature = "serde")]
```
ralexstokes, line 95: not sure what this has to do w/ `serde`? this `From` impl is just a convenience to make it easier to construct `U256` values

if this pulls in some `std` feature then I'd rather focus on handling that and not just batch into the `serde` feature

ralexstokes (commented):
this is amazing! nice work :)

I left some questions/comments on an initial pass

At a high-level, I think this direction makes sense -- the only thing I'd suggest is to take inspiration from the `serde` library which also supports std vs no-std in this way but they have a central `lib` that does the import logic once and then can simply `use lib::*;` everywhere else

see here: https://github.com/serde-rs/serde/blob/master/serde/src/lib.rs#L151

and here for usage: https://github.com/serde-rs/serde/blob/master/serde/src/ser/mod.rs#L110


I'll want to make another pass to look more deeply into the `thiserror` issue and also investigate some of the dependency changes but I like the direction this is going in and think `no-std` is a great addition to this library

.github/workflows/rust.yml
```diff
@@ -20,6 +20,8 @@ jobs:
       run: rustup update stable
     - name: Build
       run: cargo build --verbose
+    - name: Build no-std
+      run: cargo build --no-default-features --verbose
```
ralexstokes, line 24: I think we will also want to run the tests in a `no-std` mode

ralexstokes: this can be another PR but for reference this is the latest version of this CI suite I like for this set of repos:

https://github.com/ralexstokes/ethereum-consensus/blob/main/.github/workflows/rust.yml

ssz-rs/Cargo.toml
```diff
@@ -16,18 +16,25 @@ exclude = [
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["serde"]
+default = ["serde", "std"]
+std = [
+    "bitvec/std",
+    "sha2/std",
+    "lazy_static/spin",
```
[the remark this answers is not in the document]
claravanstaden: Good point, addressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/da0ebb45b1ed14d0768782564b56c2fecbd746ca.

ssz-rs/src/lib.rs
```diff
@@ -12,18 +17,16 @@ mod serde_test;
 mod uint;
 mod union;
 mod vector;
+pub mod std;
```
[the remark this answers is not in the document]
claravanstaden: The tests in `ssz-rs/examples` used the std mod, but I guess it doesn't need to. I changed it to the std lib and removed pub. https://github.com/ralexstokes/ssz-rs/pull/25/commits/5ba57306013229038ef004da29fb3111eb7bde37 and https://github.com/ralexstokes/ssz-rs/pull/25/commits/5ba57306013229038ef004da29fb3111eb7bde37.

ssz-rs/src/std.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2020 Snowfork
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE or
+// http://www.apache.org/licenses/LICENSE-2.0>. This file may not be
+// copied, modified, or distributed except according to those terms.
```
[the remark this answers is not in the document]
claravanstaden: Good point! Copy pasted from our source code. 😄 Remove in https://github.com/ralexstokes/ssz-rs/pull/25/commits/a5f17ef7bb58455db8ea95d0053b0f0d4171a158.

ssz-rs/src/uint.rs
claravanstaden: Whoops, I added this accidentally during the rebase. Removed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/1ff4b88163df1e0178af8da0fc85c4e27d55f1d1.

.github/workflows/rust.yml
claravanstaden: I updated the CI. https://github.com/ralexstokes/ssz-rs/pull/25/commits/f4cc01baf8fa43578325d49adf7cc71258b8c208

Running tests in no-std mode isn't something that really makes sense because that would mean having to remove the crates used in testing, like the `hex` crate. As far as I know running tests can still depend on std (our bridge test code also relies on std).

.github/workflows/rust.yml
```diff
@@ -15,12 +15,19 @@ jobs:
...
+      - name: Format
+        run: cargo fmt -- --check
+      - name: Run clippy
+        run: cargo clippy --all-features --verbose -- -D warnings
+      - name: Build
+        run: cargo build --all-features --verbose
+      - name: Run tests
+        run: cargo test --verbose
+      - name: Build no-std
+        run: cargo build --no-default-features --verbose
```
ralexstokes, line 32:
```suggestion
      - name: Build
        run: cargo build --all-features --verbose
      - name: Build no-std
        run: cargo build --no-default-features --verbose
      - name: Run tests
        run: cargo test --verbose
```

ssz-rs/examples/another_container.rs
```diff
@@ -1,6 +1,5 @@
 use hex;
 use ssz_rs::prelude::*;
-use std::iter::FromIterator;
```
ralexstokes, line 3: nice catch, I've removed these on `main` and you can rebase to pull in those changes

ssz-rs/src/bitlist.rs
```diff
@@ -88,7 +86,9 @@ impl<const N: usize> Bitlist<N> {
 
     fn pack_bits(&self) -> Result<Vec<u8>, MerkleizationError> {
         let mut data = vec![];
-        let _ = self.serialize_with_length(&mut data, false)?;
+        let _ = self
+            .serialize_with_length(&mut data, false)
+            .map_err(|_| MerkleizationError::SerializationError);
```
ralexstokes, line 91: this just ignores the error?

ssz-rs/src/bitvector.rs
```diff
@@ -95,7 +93,9 @@ impl<const N: usize> Bitvector<N> {
 
     fn pack_bits(&self) -> Result<Vec<u8>, MerkleizationError> {
         let mut data = vec![];
-        let _ = self.serialize(&mut data)?;
+        let _ = self
+            .serialize(&mut data)
+            .map_err(|_| MerkleizationError::SerializationError);
```
ralexstokes, line 98: same thing here?

ssz-rs/src/lib.rs
```diff
@@ -9,21 +14,19 @@ mod merkleization;
...
 #[cfg(feature = "serde")]
 mod serde_test;
+mod std;
 mod uint;
 mod union;
 mod vector;
 
-use crate::list::Error as ListError;
-use crate::vector::Error as VectorError;
+pub use crate::std::*;
```
ralexstokes, line 22: does `ssz-rs` need to expose `crate::std::*`?

it seems like no but perhaps we need to export our exact types for other no-std code to use ours?

ssz-rs/src/lib.rs
```diff
@@ -9,21 +14,19 @@ mod merkleization;
...
 pub use bitlist::Bitlist;
 pub use bitvector::Bitvector;
 pub use de::{Deserialize, DeserializeError};
 pub use list::List;
 pub use merkleization::{Context as MerkleizationContext, MerkleizationError, Merkleized, Node};
 pub use ser::{Serialize, SerializeError};
-use thiserror::Error;
 pub use uint::U256;
-pub use vector::Vector;
```
ralexstokes, line 26: why do we drop the export of `Vector` here?

ssz-rs/src/uint.rs
```diff
@@ -45,7 +44,9 @@ macro_rules! define_uint {
         impl Merkleized for $uint {
             fn hash_tree_root(&mut self) -> Result<Node, MerkleizationError> {
                 let mut root = vec![];
-                let _ = self.serialize(&mut root)?;
+                let _ = self
+                    .serialize(&mut root)
+                    .map_err(|_| MerkleizationError::SerializationError);
```
ralexstokes, line 49: same thing here w/ dropping the error checking

-- I see we do this because we lose the auto-derive `From` impl -- I'd actually prefer to keep it if we can and while it means adding more code manually we don't have to worry about correctness anywhere else -- can you add the `From` impls?

ssz-rs/examples/another_container.rs
ralexstokes: ah I see this is in a few places, sorry for extra rebase work and we will keep the others in this PR :)

ralexstokes (commented):
nice updates!

things are looking good, I left a few questions

ralexstokes:
> Or rename it something else.

I like how `serde` does it and just calls it `lib`

can you update the mod name and then the imports so that it is `crate::lib::*` etc? thank you!

ralexstokes:
also the fact that the tests are passing even though we are 'ignoring' some errors suggests the code coverage is not ideal -- I'll leave a note here to investigate further

ssz-rs/src/lib.rs
```diff
@@ -64,14 +67,13 @@ where
...
-    Deserialize(#[from] DeserializeError),
-    Merkleization(#[from] MerkleizationError),
-    List(#[from] ListError),
-    Vector(#[from] VectorError),
+    Serialize,
+    Deserialize,
+    Merkleization,
+    List,
+    Vector,
 }
```
ralexstokes, line 77: should be straightforward to add a `fmt::Display` impl to retain backwards compatibility with `thiserror` functionality

ssz-rs/examples/container_with_some_types.rs
```diff
@@ -53,15 +52,15 @@ fn main() {
...
+            eprintln!("some error encoding: {:?}", e);
             return;
         }
     };
 
     let mut restored_foo = match Foo::<4>::deserialize(&encoding) {
         Ok(value) => value,
         Err(e) => {
-            eprintln!("some error decoding: {}", e);
+            eprintln!("some error decoding: {:?}", e);
```
ralexstokes, line 63: should be able to drop this change if we add a `fmt::Display` for `SimpleSerializeError`

ssz-rs/src/lib.rs
claravanstaden: We don't need to expose it, you're right. Addressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/02961035ee736414a965e9f8fefceae01e1c98a4.

ssz-rs/src/lib.rs
claravanstaden: Added back in https://github.com/ralexstokes/ssz-rs/pull/25/commits/02961035ee736414a965e9f8fefceae01e1c98a4.

ssz-rs/src/bitlist.rs
claravanstaden: Whoops, that isn't great! Addressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/b58448c30caa3cfc15bf696c020abb01001cac89.

ssz-rs/src/bitvector.rs
claravanstaden: Addressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/b58448c30caa3cfc15bf696c020abb01001cac89.

ssz-rs/src/uint.rs
claravanstaden: Totally, much cleaner. Addressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/b58448c30caa3cfc15bf696c020abb01001cac89.

ssz-rs/examples/container_with_some_types.rs
claravanstaden: Addressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/b58448c30caa3cfc15bf696c020abb01001cac89.

claravanstaden:
> I like how serde does it and just calls it lib
>
> can you update the mod name and then the imports so that it is crate::lib::* etc? thank you!

Done in https://github.com/ralexstokes/ssz-rs/pull/25/commits/d77632666961f66c49ff0ea8eaf5c21ecaf46dac. I removed the `std` mod and created a `lib` mod in `lib.rs`, since lib.rs exists already.

ssz-rs/src/lib.rs
claravanstaden: Totally, good point. Implemented in https://github.com/ralexstokes/ssz-rs/pull/25/commits/ea113b48f3ad134603f11c9be9f5fc696d30c259.

claravanstaden:
> also the fact that the tests are passing even though we are 'ignoring' some errors suggests the code coverage is not ideal -- I'll leave a note here to investigate further

This is a little worrying. The dropped error check I introduced was in pack_bytes, used by `hash_tree_root`, which the individual mod tests don't cover. But it is still concerning that the container tests didn't pick them up.

I had a quick look but must admit I couldn't quickly see how to add test cases that would trigger those error conditions.

dharjeezy:
Is this still being worked on? Is there any way i can contribute to this?

claravanstaden:
> Is this still being worked on? Is there any way i can contribute to this?

@dharjeezy it's ready for review! @ralexstokes will probably get to it as soon as he can. 😄 You are welcome to also review and test if you are keen.

.github/workflows/rust.yml
```diff
@@ -15,12 +15,19 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
```
ralexstokes, line 18: I'm working to tidy up the CI across this family of repos, we will need to rebase that once it is in place; trying to finish in the next 6 hours or so

ssz-rs/src/bitlist.rs
```diff
@@ -87,7 +86,7 @@ impl<const N: usize> Bitlist<N> {
 
     fn pack_bits(&self) -> Result<Vec<u8>, MerkleizationError> {
         let mut data = vec![];
-        let _ = self.serialize_with_length(&mut data, false)?;
+        let _ = self.serialize_with_length(&mut data, false);
```
ralexstokes, line 89: we want to keep this right?

ssz-rs/src/bitvector.rs
```diff
@@ -94,7 +93,7 @@ impl<const N: usize> Bitvector<N> {
 
     fn pack_bits(&self) -> Result<Vec<u8>, MerkleizationError> {
         let mut data = vec![];
-        let _ = self.serialize(&mut data)?;
+        let _ = self.serialize(&mut data);
```
ralexstokes, line 96: keep this `?`?

ssz-rs/src/list.rs
```diff
@@ -294,7 +297,7 @@ pub struct IterMut<'a, T, const N: usize>
 where
     T: SimpleSerialize,
 {
-    inner: Enumerate<slice::IterMut<'a, T>>,
+    inner: Enumerate<crate::lib::IterMut<'a, T>>,
```
ralexstokes, line 300: if we did `use crate::lib::*;` up above, `IterMut` should be in scope, right?

ssz-rs/src/merkleization/mod.rs
```diff
@@ -21,15 +19,35 @@ pub trait Merkleized {
...
-    InputExceedsLimit(usize),
+    SerializationError,           // failed to serialize value
+    PartialChunk(Vec<u8>, usize), // cannot merkleize a partial chunk of length {1} (data: {0:?})
+    InputExceedsLimit(usize),     // cannot merkleize data that exceeds the declared limit {0}
+}
+
+impl From<SerializeError> for MerkleizationError {
+    fn from(_: SerializeError) -> Self {
+        return MerkleizationError::SerializationError
+    }
```
ralexstokes, line 32: can we not wrap the serialization error? even if we have to string-ify or something?

ssz-rs/src/ser.rs
```diff
@@ -1,19 +1,37 @@
...
 pub enum SerializeError {
-    #[error("the encoded length is {0} which exceeds the maximum length {MAXIMUM_LENGTH}")]
-    MaximumEncodedLengthExceeded(usize),
-    #[error("the type for this value has a bound of {bound} but the value has {len} elements")]
-    TypeBoundsViolated { bound: usize, len: usize },
-    #[error("the type for this value has an illegal bound of {bound}")]
-    IllegalType { bound: usize },
+    MaximumEncodedLengthExceeded(usize), // the encoded length is {0} which exceeds the maximum length {MAXIMUM_LENGTH}
+    TypeBoundsViolated { bound: usize, len: usize }, // the type for this value has a bound of {bound} but the value has {len} elements"
+    IllegalType { bound: usize }, // the type for this value has an illegal bound of {bound}
```
ralexstokes, line 12: prolly drop these comments, esp if we are just repeating what the `fmt::Display` is

ralexstokes (commented):
a few comments to resolve, but I think we are only a round or two of back/forth before we can merge :)

ssz-rs/src/list.rs
claravanstaden: It looks like `IterMut` conflicts with the struct definition `pub struct IterMut<'a, T, const N: usize>`, so I have had to specify that it is the `IterMut` from the lib.

.github/workflows/rust.yml
claravanstaden: Great! I rebased and adding building no-std. https://github.com/ralexstokes/ssz-rs/pull/25/files#diff-73e17259d77e5fbef83b2bdbbe4dc40a912f807472287f7f45b77e0cbf78792dR31

ssz-rs/src/bitlist.rs
claravanstaden: Adressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/438a003498f6a5be024b30a8137c8673eb05895a.

ssz-rs/src/bitvector.rs
claravanstaden: Addressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/438a003498f6a5be024b30a8137c8673eb05895a.

ssz-rs/src/merkleization/mod.rs
claravanstaden: Sure, good idea! Addressed in https://github.com/ralexstokes/ssz-rs/pull/25/commits/438a003498f6a5be024b30a8137c8673eb05895a.

ssz-rs/src/ser.rs
claravanstaden: Good point, removed. https://github.com/ralexstokes/ssz-rs/pull/25/commits/438a003498f6a5be024b30a8137c8673eb05895a

claravanstaden:
@ralexstokes thanks for the review! 😄 Comments actioned.

ssz-rs/Cargo.toml
```diff
@@ -16,18 +16,24 @@ exclude = [
...
 
 [features]
-default = ["serde"]
+default = ["serde", "std"]
+std = [
+    "bitvec/std",
+    "sha2/std",
+    "num-bigint/std",
+    "serde/std"
+]
```
ralexstokes, line 25: unless there is some collision, it seems better to just pull in the `default` feature set for each crate

```suggestion
std = [
    "bitvec/default",
    "sha2/default",
    "lazy_static/default",
    "num-bigint/default",
]
```

I also think we can drop `serde` here? may have to try to build w/o it and see what happens

ssz-rs/Cargo.toml
```diff
@@ -16,18 +16,24 @@ exclude = [
...
 serde = ["dep:serde", "dep:hex"]
 
 [dependencies]
 thiserror = "1.0.25"
-bitvec = "1.0.0"
+bitvec = { version = "1.0.0", default-features = false, features=["alloc"]}
```
ralexstokes, line 30: if we go w/ the `default` features, then we can do this:
```suggestion
bitvec = { version = "1.0.0", default-features = false }
```

ssz-rs/Cargo.toml
```diff
@@ -16,18 +16,24 @@ exclude = [
...
 ssz-rs-derive = { path = "../ssz-rs-derive"}
-sha2 = "0.9.8"
-lazy_static = "1.4.0"
+sha2 = { version ="0.9.8", default-features = false}
+lazy_static = { version = "1.4.0", features = ["spin_no_std"]}
```
ralexstokes, line 33: I think we can try this:

```suggestion
lazy_static = { version = "1.4.0", default-features = false, features = ["spin_no_std"]}
```

ralexstokes (commented):
reviewed the Cargo manifest and I think we can tighten it up some

I'd like to get to a point where the default thing is the status quo and only if you do `--no-default-features` will you get a `no_std` compatible library

ssz-rs/src/lib.rs
```diff
@@ -13,18 +18,42 @@ mod uint;
 mod union;
 mod vector;
 
-use crate::list::Error as ListError;
-use crate::vector::Error as VectorError;
+use crate::lib::*;
```
ralexstokes, line 21: I admit this is nitpicky, but...

I don't love tossing all of the random symbols from `crate::lib::*;`  into the `lib` module namespace

but it looks like we did this for the fns `serialize` and `deserialize`, along with the `SimpleSerializeError`

to resolve this, I'd suggest making a new file `utils.rs` with `serialize and `deserialize` (and the required imports)

along with moving the error type to a new file `error.rs` (and while we are at it just calling it `Error`)

ssz-rs/src/lib.rs
```diff
@@ -64,14 +93,27 @@ where
...
+impl Display for SimpleSerializeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        match self {
+            SimpleSerializeError::Serialize(error) => write!(f, "serialize error: {}", error),
+            SimpleSerializeError::Deserialize(error) => write!(f, "deserialize error: {}", error),
+            SimpleSerializeError::Merkleization(error) => {
+                write!(f, "merkleization error: {}", error)
+            }
+            SimpleSerializeError::List(error) => write!(f, "list error: {}", error),
+            SimpleSerializeError::Vector(error) => write!(f, "vector error: {}", error),
```
ralexstokes, line 114: looks like we can replace all of these with `write!(f, "{error}"),

to retain the existing error formatting

ssz-rs/src/list.rs
```diff
@@ -1,26 +1,33 @@
...
-    #[error("{provided} elements given that exceeds the list bound of {expected}")]
+use serde::ser::SerializeSeq;
+
+#[derive(Debug)]
+pub enum ListError {
     IncorrectLength { expected: usize, provided: usize },
 }
 
+impl Display for ListError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
```
ralexstokes, line 20: same thing on `core::fmt::Result` -- if we need to do something more verbose, I would actually just leave these all as is!

ssz-rs/src/list.rs
```diff
@@ -1,26 +1,33 @@
...
+        match *self {
```
ralexstokes, line 21: do we need to deref `self` here?

ssz-rs/src/merkleization/mod.rs
```diff
@@ -2,14 +2,12 @@ mod cache;
...
+use crate::lib::*;
+use crate::ser::Serialize;
 use lazy_static::lazy_static;
 use sha2::{Digest, Sha256};
-use std::cmp::Ordering;
-use std::fmt::Debug;
-use std::ops::Index;
-use thiserror::Error;
 
+use crate::SerializeError;
```
ralexstokes, line 10: why grab this from the top-level, instead of leaving the import as it was?

ssz-rs/src/merkleization/mod.rs
```diff
@@ -21,17 +19,39 @@ pub trait Merkleized {
...
 
+impl From<SerializeError> for MerkleizationError {
+    fn from(e: SerializeError) -> Self {
+        MerkleizationError::SerializationError(e)
+    }
+}
+
+impl Display for MerkleizationError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        match self {
```
ralexstokes, line 37: same thing on the prefix "the value could not be merkleized:"

ssz-rs/src/merkleization/mod.rs
```diff
@@ -21,17 +19,39 @@ pub trait Merkleized {
...
+            MerkleizationError::SerializationError(e) => {
+                write!(f, "failed to serialize value: {}", e)
+            }
+            MerkleizationError::PartialChunk(chunk, size) => write!(
+                f,
+                "cannot merkleize a partial chunk of length {} (data: {:?}",
```
ralexstokes, line 43:
```suggestion
                "cannot merkleize a partial chunk of length {} (data: {:?})",
```

ssz-rs/src/merkleization/mod.rs
```diff
@@ -21,17 +19,39 @@ pub trait Merkleized {
...
+    fn from(e: SerializeError) -> Self {
+        MerkleizationError::SerializationError(e)
+    }
+}
+
+impl Display for MerkleizationError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        match self {
+            MerkleizationError::SerializationError(e) => {
+                write!(f, "failed to serialize value: {}", e)
```
ralexstokes, line 39: if we look at the inner error message, we don't need to add any additional text here
```suggestion
                write!(f, "{e}")
```

ssz-rs/src/ser.rs
```diff
@@ -1,21 +1,39 @@
...
     MaximumEncodedLengthExceeded(usize),
-    #[error("the type for this value has a bound of {bound} but the value has {len} elements")]
     TypeBoundsViolated { bound: usize, len: usize },
-    #[error("the type for this value has an illegal bound of {bound}")]
     IllegalType { bound: usize },
 }
 
+impl Display for SerializeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        match *self {
```
ralexstokes, line 17: same thing on prefix here

ssz-rs/src/ser.rs
```diff
@@ -1,21 +1,39 @@
...
-    #[error("the encoded length is {0} which exceeds the maximum length {MAXIMUM_LENGTH}")]
     MaximumEncodedLengthExceeded(usize),
-    #[error("the type for this value has a bound of {bound} but the value has {len} elements")]
     TypeBoundsViolated { bound: usize, len: usize },
-    #[error("the type for this value has an illegal bound of {bound}")]
     IllegalType { bound: usize },
 }
 
+impl Display for SerializeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
```
ralexstokes, line 16: can just say `fmt::Result` here?

ssz-rs/src/vector.rs
```diff
@@ -1,22 +1,17 @@
 use crate::de::{deserialize_homogeneous_composite, Deserialize, DeserializeError};
+use crate::lib::*;
 use crate::merkleization::{
     merkleize, pack, MerkleCache, MerkleizationError, Merkleized, Node, BYTES_PER_CHUNK,
 };
-use crate::ser::{serialize_composite, Serialize, SerializeError};
-use crate::{SimpleSerialize, SimpleSerializeError, Sized};
+use crate::ser::{serialize_composite, Serialize};
+use crate::{SerializeError, SimpleSerialize, SimpleSerializeError, Sized};
```
ralexstokes, line 7: don't see reason to change import path on `SerializeError`...

ralexstokes (commented):
another round of questions/comments :)

ssz-rs/src/ser.rs
ralexstokes: ah also, do we need to deref self?

ssz-rs/src/de.rs
```diff
@@ -1,24 +1,40 @@
...
-    IOError(#[from] std::io::Error),
-    #[error("the type for this value has a bound of {bound} but the value has {len} elements")]
+    IOError,
     TypeBoundsViolated { bound: usize, len: usize },
-    #[error("the type for this value has an illegal bound of {bound}")]
     IllegalType { bound: usize },
 }
 
+impl Display for DeserializeError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
```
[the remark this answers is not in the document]
seunlanlege: I would actually argue that its best to keep libraries platform independent by using core especially when they do not depend on any OS specific APIs

ssz-rs/src/lib.rs
ralexstokes: ok, I made these changes on `main`, if you rebase you'll see what needs to happen (should just be a few `crate::lib::*` imports)

ssz-rs/src/de.rs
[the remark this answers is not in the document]
seunlanlege: Yeah it's done conditionally, but I meant you could get away with making the entire lib no_std by default with no linking to the standard library.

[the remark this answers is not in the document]
ralexstokes: if I'm following you then I hear what you are saying but this library certain will pull in other things from `std` (e.g. `Vec`) and so it is not easy to pursue `no_std` by default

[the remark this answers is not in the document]
seunlanlege: So unless you depend on some OS-specific API, you can just use the `alloc` crate for heap allocated types.

https://doc.rust-lang.org/alloc/

std actually re-exports alloc.

ssz-rs/src/lib.rs
```diff
@@ -1,3 +1,8 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+#[cfg(not(feature = "std"))]
+extern crate alloc;
```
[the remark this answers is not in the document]
claravanstaden: I do think we need this. If I remove this, I get the following error:

```
error[E0433]: failed to resolve: use of undeclared crate or module `alloc`
  --> ssz-rs/src/lib.rs:49:13
   |
49 |     pub use alloc::{format, string::String, vec, vec::Vec};
   |             ^^^^^ use of undeclared crate or module `alloc`
   |
   = help: add `extern crate alloc` to use the `alloc` crate
```

ssz-rs/src/de.rs
```diff
@@ -1,24 +1,40 @@
...
+        match *self {
```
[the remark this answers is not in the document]
claravanstaden: Sure, agreed! I couldn't figure out how to print out the `value` in question, like `this error`. Let me know if this is OK: https://github.com/ralexstokes/ssz-rs/pull/25/commits/6a4659db85d81bc1456a4357e18dfc11e636d247#diff-4f175038d54055a59b936ca119f3635102a314975c7369d2235e31c7445586e1R16

ssz-rs/src/lib.rs
```diff
@@ -13,18 +18,42 @@ mod uint;
 mod union;
 mod vector;
 
-use crate::list::Error as ListError;
-use crate::vector::Error as VectorError;
+use crate::lib::*;
+use crate::list::ListError;
+use crate::vector::VectorError;
```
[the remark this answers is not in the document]
claravanstaden: Totally, I think I changed this when I was still figuring out how Rust mod encapsulation works and thought there might be name collision. Reverted.

ssz-rs/src/lib.rs
claravanstaden: Much cleaner, thanks. I adjusted my changes accordingly.

ssz-rs/src/lib.rs
claravanstaden: I ended up removing these as they aren't necessary anymore.

ssz-rs/src/list.rs
claravanstaden: Nice catch, a quick fix.

ssz-rs/src/list.rs
claravanstaden: We don't, removed.

ssz-rs/src/merkleization/mod.rs
claravanstaden: Reverted.

ssz-rs/src/vector.rs
claravanstaden: Agreed, reverted.

ssz-rs/Cargo.toml
claravanstaden: This doesn't work because if feature `std` isn't enabled, we still need the `alloc` feature of `bitvec`.

ssz-rs/Cargo.toml
claravanstaden: Unfortunately `lazy-static` doesn't have a `default` feature: https://github.com/rust-lang-nursery/lazy-static.rs/blob/master/Cargo.toml

But the rest works nicely, I could drop `serde` here, as you suggested.

claravanstaden:
@ralexstokes thanks for the review! I've made the requested changes.

The one thing I am wondering about is how the new error mod fits into exposing child errors, if the crate user wants to check specific data structures like `list::Error::IncorrectLength`, like this: https://github.com/claravanstaden/use-ssz/blob/main/src/main.rs#L14. Or even just wants to check if Error is a List error:https://github.com/claravanstaden/use-ssz/blob/main/src/main.rs#L18 This doesn't seem to work:

```
error[E0308]: mismatched types
  --> src/main.rs:13:17
   |
12 |             match err {
   |                   --- this expression has type `ssz_rs::list::Error`
13 |                 Error::List(e) => println!("list error: {e}")
   |                 ^^^^^^^^^^^^^^ expected enum `ssz_rs::list::Error`, found enum `ssz_rs::Error`
```

Perhaps some explicit mapping between the new error module and child crates are needed for this to work?

Along the same line of thought, should we expose `Error as ListError` [here](https://github.com/ralexstokes/ssz-rs/blob/main/ssz-rs/src/lib.rs#L22)? So that the user can match the kinds of errors (I mean in this case there is only one type of error `IncorrectLength`, so maybe it doesn't matter as much in this case, but just wondering about your thoughts on this).

ssz-rs/src/de.rs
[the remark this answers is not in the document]
ralexstokes: which is fine but it seems unnecessarily surprising to use an `alloc` import when something is in `std` as most Rust users will just be embedded in `std` by default

this seems like the kind of thing where you need a very specific reason to deviate from the community-accepted norm/idiom

ssz-rs/Cargo.toml
ralexstokes: ah then I think we can just do `"lazy_static"` and then keep it "default-features false" and use the `spin_no_std` feature for our `no_std` setting

ssz-rs/src/de.rs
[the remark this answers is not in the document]
ralexstokes: hmm ok I apparently made up a feature of `thiserror`!

I thought that putting a `error` attribute on the outer part of the type would *prefix* all the inner messages with that message

e.g.
```rust
#[derive(Debug, Error)]
#[error("something {0}")]
enum Foo {
  #[error("is wrong with A")]
  A(u8),
  #[error("is wrong with B")]
  B(u8),
}
// the message would be "something is wrong with {A,B}"
```

but it looks like these placements are exclusionary w/ the more precise message just taking precedence over the general message


I will go fix this on `main` and then can revisit this; sorry for the confusion!

ssz-rs/src/lib.rs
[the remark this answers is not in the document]
ralexstokes: thanks for checking; this is how you do this and it is a bit different than other ways to require crates bc of its interaction with the compiler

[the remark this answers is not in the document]
claravanstaden: Cool, that makes sense yeah!

ssz-rs/src/de.rs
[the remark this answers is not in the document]
claravanstaden: Cool cool, will rebase once merged. 😄 That would have been more useful behaviour by `this error` tbh.

ralexstokes:
@claravanstaden I refactored the errors so this is no longer a concern; I haven't looked at the previous setup but the intention is to provide the user the ability to match on any 'internal' error to this library

that being said, there is the "top-level" `Error` that aggregates everything and serves as the entry point for most users of the library (renamed to `SimpleSerializeError` in the `prelude` so that `use ssz_rs::prelude::*;` doesn't try to clobber any `Error` symbol the user may also have in scope)

ralexstokes (commented):
nice updates! looks like the only real thing left is to fix the errors

I'm happy to do this assuming I can push to your branch so let me know which you prefer :)

Otherwise, we have whittled the PR down to just the two blocking things for `no-std`, which would be the error situation (with `thiserror` incompatibility) and conversion of imports so we can pick std vs no-std variants depending on the library feature flag. Nice :)

claravanstaden:
Awesome @ralexstokes! I'm happy if you make the changes, I've given you write access on the `Snowfork/ssz_rs` repo. If you don't make the changes today, I'll make them tomorrow.

ssz-rs/Cargo.toml
ralexstokes: it turns out that `lazy_static` is not easy to integrate in an additive way, so I just dropped it and added a build script to generate the required data at compile time, see #42 for more info

ssz-rs/src/merkleization/mod.rs
```diff
@@ -20,14 +17,29 @@ pub trait Merkleized {
...
-    #[error("failed to serialize value: {0}")]
-    SerializationError(#[from] SerializeError),
-    #[error("cannot merkleize data that exceeds the declared limit {0}")]
+    SerializationError(SerializeError),
     InputExceedsLimit(usize),
 }
 
+impl From<SerializeError> for MerkleizationError {
+    fn from(e: SerializeError) -> Self {
+        MerkleizationError::SerializationError(e)
```
ralexstokes, line 28:
```suggestion
    fn from(err: SerializeError) -> Self {
        MerkleizationError::SerializationError(err)
```

Issue #27, opened 2022-11-18
overflow when compiling for 32 bit architectures


ncitron:
When compiling to `wasm32-unknown-unknown`, [this line](https://github.com/ralexstokes/ssz-rs/blob/main/ssz-rs/src/ser.rs#L6) overflows, since 2^32 cannot fit inside of `usize`. I hacked my way around this by just hard coding it to 2^32-1, but I imagine that is going to break something elsewhere. Guessing the better move would be to convert to `u64`.

PR #28, opened 2022-11-20
fix: avoid overflow on 32 bit architectures

ncitron:
addresses #27

### DIFF
--- a/ssz-rs/src/ser.rs
+++ b/ssz-rs/src/ser.rs
@@ -3,7 +3,7 @@ use thiserror::Error;
 
 // NOTE: if this is changed, go change in `ssz_derive` as well!
 pub const BYTES_PER_LENGTH_OFFSET: usize = 4;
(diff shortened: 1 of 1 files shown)


claravanstaden:
I also addressed this in https://github.com/ralexstokes/ssz-rs/pull/25/files#diff-97dff6cb7d3f1bea2c98e61d85eeaab0dfc941e86dbc3679aa69a51d528c8801R6. 😃 We can probably merge this one first @ralexstokes, nice and small.

ralexstokes (approved):
this works for me!

I don't really see why but if there is ever any reason why we want to optimize out the extra 4 bytes of overhead we can just hardcode the limit directly

also it looks like there are several clippy lints added in the latest rust version that is failing the build -- I can fix them and then we can rebase this on top of that

ralexstokes:
@ncitron if you rebase from `main` this should unblock the CI

ncitron:
Rebased

PR #32, opened 2022-12-29
ssz merkle multi prover/verifier

### LINKED ISSUE #35: support Merkle multi proofs
support proving and verification of Merkle (multi)proofs

seunlanlege:
Fixes #35.

I'm currently trying to add support for generating ssz merkle multi proofs, follwing the ssz spec. I'm breaking down the work into multiple commits:

Fist I added the neccessary runtime reflection code needed for [`get_generalized_index`](https://github.com/ethereum/consensus-specs/blob/dev/ssz/merkle-proofs.md#ssz-object-to-index) for `SszObject`s.

Then I introduced the proof generation code, based on this runtime reflection.

Apologies for the cargo fmt changes, i don't like to format my code by hand.

edit(ralexstokes): link issue


ssz-rs-derive/src/lib.rs
```diff
@@ -562,6 +645,22 @@ pub fn derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
         }
 
         #impl_impl ssz_rs::SimpleSerialize for #name_impl {}
+
+        #fields_inspect_impl
+
+        #impl_impl ssz_rs::SszReflect for #name_impl {
+            fn ssz_type_class(&self) -> ssz_rs::SszTypeClass {
+                ssz_rs::SszTypeClass::Container
```
seunlanlege, line 653: should either be a container, enum or union

ssz-rs/Cargo.toml
```diff
@@ -31,7 +31,9 @@ ssz-rs-derive = { path = "../ssz-rs-derive"}
 sha2 = { version ="0.9.8", default-features = false}
 serde = { version = "1.0", features = ["derive"], optional = true }
 hex = {version = "0.4.3", optional = true }
-num-bigint = { version ="0.4.3", default-features = false}
+as-any = "0.3.0"
```
seunlanlege, line 34: this crate is `no_std` by default

ssz-rs/src/list.rs
```diff
@@ -290,6 +291,23 @@ where
...
+impl<T, const N: usize> SszReflect for List<T, N>
+where
+    T: SimpleSerialize + SszReflect + AsAny,
+{
+    fn ssz_type_class(&self) -> SszTypeClass {
+        SszTypeClass::Elements(ElementsType::List)
+    }
+
+    fn list_elem_type(&self) -> Option<&dyn SszReflect> {
+        Some(self.index(0))
```
seunlanlege, line 303: so this is going to be a problem for sure, as this won't work for empty lists

ssz-rs/src/union.rs
```diff
@@ -70,7 +70,14 @@ where
     }
 }
 
-impl<T> SimpleSerialize for Option<T> where T: SimpleSerialize {}
+impl<T> SszReflect for Option<T>
+where
+    T: SimpleSerialize + SszReflect,
+{
+    fn ssz_type_class(&self) -> SszTypeClass {
+        SszTypeClass::Container // really?
```
seunlanlege, line 78: Should be union

ssz-rs/src/list.rs
ralexstokes: return `None` if empty? I don't follow exactly what is going on here yet

seunlanlege:
@ralexstokes still interested in this?

ralexstokes:
@seunlanlege I am, although I haven't had time to digest all of the reflective code here -- I am wondering if there is a way to move it to compile-time; pushing most of the logic in this PR to some proc macro and then only doing the bare minimum at runtime, e.g. interpreting GeneralizedIndices and returning parts of the backing Merkle tree

ralexstokes:
hey @seunlanlege thanks for this! I'm going to close this PR as I have settled on the compile-time route for generalized index computation currently on `main`, and I've started working on a proving scheme in #131 to support basic (single) merkle proofs.

if you wanted to look at multiproof generation + verification following this work, would be happy to discuss further :)

seunlanlege:
> hey @seunlanlege thanks for this! I'm going to close this PR as I have settled on the compile-time route for generalized index computation currently on `main`, and I've started working on a proving scheme in #131 to support basic (single) merkle proofs.
> 
> 
> 
> if you wanted to look at multiproof generation + verification following this work, would be happy to discuss further :)

Interesting stuff, glad you were able to finally figure it out. Unfortunately I won't have the bandwidth to contribute further at this time. We're already using my runtime-indexing approach in our beacon chain consensus prover for [hyperbridge](https://github.com/polytope-labs/hyperbridge/blob/main/modules/consensus/sync-committee/prover/src/lib.rs#L441).

It Works pretty well, and it's already been live on our testnet since last year infact we're just about to go through audits for mainnet. We will definitely look into maybe switching to the compile time version once it's ready :D

PR #42, opened 2023-01-23
moves computation of merkle "zero hashes" to build time

ralexstokes:
- in order to drop use of `lazy_static` dep which is not easy to use in `no-std` context


ralexstokes:
see here for some more context around using `lazy_static` in `no-std` contexts and how it is tricky...

https://github.com/entropyxyz/crypto-primes/issues/7#issue-1515227220

Issue #22, opened 2022-08-23
enhance indexing and arithmetic safety


ralexstokes:
analogues of these:

https://github.com/ralexstokes/ethereum-consensus/issues/151

https://github.com/ralexstokes/ethereum-consensus/issues/152

for this repo :)

claravanstaden:
Hey @ralexstokes! As part of our production readiness of https://github.com/Snowfork/snowbridge I will be taking a look at these.

ralexstokes:
@0xTylerHolmes found a panic here :)

```rust
use ssz_rs::prelude::*;

type Bar = Vector<u8, 96>;

#[derive(PartialEq, Eq, Debug, Default, Clone, SimpleSerialize)]
pub struct Foo {
    pub bar: Bar,
}

fn main() {
    let data = vec![];
    let foo = Foo::deserialize(&data);
    dbg!(foo);
}
```

and what fails is this: https://github.com/ralexstokes/ssz-rs/blob/main/ssz-rs-derive/src/lib.rs#L202 where the range on `encoding` hits out of bounds

instead, we should use https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get

ralexstokes:
this should be addressed by all of the work linked here once #57 is merged

great work everyone and thank you all SO much for improving this library

PR #54, opened 2023-04-16
add potential fix for OOR access on Vector

0xTylerHolmes:
Unlike the issue this does not use the safe vec::get method, but does emit an already defined DeserializeError.  I am open to other options for the easiest way to fix this. If this is a desired approach I can add some other guards where they are needed.

### DIFF
--- a/ssz-rs-derive/src/lib.rs
+++ b/ssz-rs-derive/src/lib.rs
@@ -199,12 +199,11 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
                         } else {
                             let encoded_length = <#field_type>::size_hint();
                             let end = start + encoded_length;
-                            if end > encoding.len() {
-                                return Err(ssz_rs::DeserializeError::ExpectedFurtherInput {
-                                provided: encoding.len(),
-                                expected: end,
-                            })};
-                            let result = <#field_type>::deserialize(&encoding[start..end])?;
+                            let enc = &encoding.get(start..end).ok_or(DeserializeError::InvalidRange{
+                                range: start..end,
+                                buffer_length:encoding.len(),
+                            })?;
+                            let result = <BLSSignature>::deserialize(&enc)?;
                             container.#field_name = result;
                             encoded_length
                         };
--- a/ssz-rs-derive/src/lib.rs
+++ b/ssz-rs-derive/src/lib.rs
@@ -192,7 +192,7 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
                     Some(field_name) => quote_spanned! { f.span() =>
                         let bytes_read = if <#field_type>::is_variable_size() {
                             let end = start + #BYTES_PER_LENGTH_OFFSET;
-                            let target = encoding.get(start..end).ok_or(
+                            let target = encoding.get(start..end).ok_or_else(||
                                 ssz_rs::DeserializeError::ExpectedFurtherInput {
                                     provided: encoding.len() - start,
                                     expected: #BYTES_PER_LENGTH_OFFSET,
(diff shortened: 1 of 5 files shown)


0xTylerHolmes:
Note this is in reference to  #22

0xTylerHolmes:
In reference to #22 I changed it to instead use the .get() method on encoding. To my knowledge there isn't a good way to get the exact reason of the failure so I just added a generic [DeserializeError::InvalidRange{range: Range<usize>, buffer_length: usize},](https://github.com/0xTylerHolmes/ssz-rs/blob/oor-vector/ssz-rs/src/de.rs#L28)

ssz-rs-derive/src/lib.rs
```diff
@@ -192,18 +192,26 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
...
                         let bytes_read = if <#field_type>::is_variable_size() {
                             let end = start + #BYTES_PER_LENGTH_OFFSET;
-                            let next_offset = u32::deserialize(&encoding[start..end])?;
+                            let next_offset = u32::deserialize(
+                                encoding.get(start..end)
+                                .ok_or(DeserializeError::InvalidRange {
+                                    range: start..end,
+                                    buffer_length:encoding.len(),
+                                })?
+                            )?;
```
ralexstokes, line 201:
```suggestion
                            let target = encoding.get(start..end).ok_or(ssz_rs::DeserializeError::ExpectedFurtherInput{
                                provided: encoding.len(),
                                expected: #BYTES_PER_LENGTH_OFFSET,
                            })?;
                            let next_offset = u32::deserialize(target)?;
```

ralexstokes (commented):
nice! let's unpack the code a bit and I think `DeserializeError::ExpectedFurtherInput` makes more sense here -- what do you think?

ssz-rs/src/de.rs
```diff
@@ -48,6 +51,7 @@ impl Display for DeserializeError {
             ),
             DeserializeError::InvalidInstance(err) => write!(f, "invalid instance: {err}"),
             DeserializeError::InvalidType(err) => write!(f, "invalid type: {err}"),
+            DeserializeError::InvalidRange{range, buffer_length} => write!(f, "failure on getting range: {range:#?}, for length: {buffer_length}"),
```
ralexstokes, line 54: if we just go w/ `ExpectedFurtherInput` we can drop all the changes to this

0xTylerHolmes: Alright Il make the change and add some tests.

ssz-rs/tests/containers.rs
```diff
@@ -21494,6 +21494,14 @@ fn test_containers_bits_struct_one_chaos_2() {
     assert_eq!(root, expected_root);
 }
 
+#[test]
+fn test_containers_not_enough_input() {
+    type Goal = Vector<u8, 6>;
+    let source = vec![];
+    let result = Goal::deserialize(&source);
```
ralexstokes, line 21501:
```suggestion
    let source = vec![];
    let result = Vector::<u8, 6>::deserialize(&source);
```

ssz-rs-derive/src/lib.rs
```diff
@@ -192,18 +192,26 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
...
-                                buffer_length:encoding.len(),
-                            })?;
-                            let result = <#field_type>::deserialize(&enc)?;
+                            let result = <#field_type>::deserialize(
+                                encoding.get(start..end)
+                                .ok_or(DeserializeError::InvalidRange {
+                                    range: start..end,
+                                    buffer_length:encoding.len(),
+                                })?
+                            )?;
```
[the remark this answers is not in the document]
0xTylerHolmes: I agree, I think these are more simple solutions. I like breaking up the fetching and deserializing as well.

PR #56, opened 2023-05-17
Remove panic defaults in List and Vector

doubledup:
These panic cases shouldn't have come up in practice, but removing them from the source ensures they can't.

Also did a pass over the other `default` implementations, but there aren't any explicit panics and the functions used look safe. An exception is `BitVector` that has an `assert!` for `N > 0` - I've added the same assert to `Vector`. An initial attempt to move this to a compile-time check with static_assertions gave some errors I couldn't easily interpret, so left it for now.


ssz-rs/src/vector.rs
```diff
@@ -118,12 +118,13 @@ where
     T: SimpleSerialize + Default + Clone,
 {
     fn default() -> Self {
+        // TODO: statically assert that N != 0, possibly with
+        // https://crates.io/crates/static_assertions
+        assert!(N != 0);
```
doubledup, line 123: I've used `N != 0` here, but `Bitvector` uses `N > 0` - is that the preferred style?

vgeddes: I think `N > 0` reads better too.

Also, I've previously tried to use `const_assert!` with const generics, and unfortunately that wasn't supported. So we can remove the `TODO`, I think.

ssz-rs/src/list.rs
```diff
@@ -92,12 +92,7 @@ where
     T: SimpleSerialize + Default,
 {
     fn default() -> Self {
-        let data = vec![];
-        match Self::try_from(data) {
-            Ok(result) => result,
-            // TODO: not ideal to panic here...
-            Err((_, err)) => panic!("{err}"),
-        }
+        Self { data: vec![], cache: MerkleCache::with_leaves(0) }
```
ralexstokes, line 95: this works! I can pull into #55

my reasoning there was to keep the symmetry with the one "entry point" for all `List` types in the `try_from` impl but its fine to inline this here

ralexstokes (commented):
addressed this as well in #55 which I intend to merge in the next few days

I'm happy to take these approaches as well, and ill just add this to the v1 milestone for now and we can figure out how to handle this week

ssz-rs/src/vector.rs
```diff
@@ -118,12 +118,11 @@ where
     T: SimpleSerialize + Default + Clone,
 {
     fn default() -> Self {
+        assert!(N > 0);
+
         let data = vec![T::default(); N];
-        match data.try_into() {
```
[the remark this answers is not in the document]
doubledup: I'm happy with either way too. The panic case can't come up anyway because `data.len() == N`. Removing the panic case makes this clearer when scanning the code, but then we lose the elegance of using `try_from` everywhere.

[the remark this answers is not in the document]
doubledup: If we keep `try_from`, we can shorten it with `map_err`: `data.try_into().map_err(|(_, err)| err).unwrap()`

doubledup:
I've gone with `try_from`, `unwrap` and a safety comment for now. This way we keep the single entry point and `unwrap` is addressed in the same way as the other uses.

As an aside, are a `List[type, 0]` and `Bitlist[0]` valid? [The spec](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md) makes this exclusion for the vector types, but not the lists. Seems like a reasonable exclusion because the lists would have to be empty.

ralexstokes:
> As an aside, are a List[type, 0] and Bitlist[0] valid? 

I ran into the same question the other day haha

the spec says they are valid via omission and while they are strange types I don't see a problem w/ them

ssz-rs/src/vector.rs
```diff
@@ -118,14 +118,13 @@ where
...
-            Ok(result) => result,
-            // SAFETY: panic
-            // ideally we will not panic here but currently there is no way
-            // to enforce statically that `N` is non-zero with const generics
-            Err((_, err)) => panic!("{err}"),
-        }
+        // SAFETY: panic can't happen because data.len() == N != 0; qed
+        // ideally we will not panic here but currently there is no way
+        // to enforce statically that `N` is non-zero with const generics
```
ralexstokes, line 126:
```suggestion
```

ssz-rs/src/vector.rs
```diff
@@ -118,14 +118,13 @@ where
     T: SimpleSerialize + Default + Clone,
 {
     fn default() -> Self {
```
ralexstokes, line 120:
```suggestion
    fn default() -> Self {
        // SAFETY: ideally we will not panic here but currently there is no way
        // to enforce statically that `N` is non-zero with const generics
```

ssz-rs/src/vector.rs
```diff
@@ -118,14 +118,13 @@ where
...
-            Ok(result) => result,
-            // SAFETY: panic
-            // ideally we will not panic here but currently there is no way
-            // to enforce statically that `N` is non-zero with const generics
-            Err((_, err)) => panic!("{err}"),
-        }
+        // SAFETY: panic can't happen because data.len() == N != 0; qed
+        // ideally we will not panic here but currently there is no way
+        // to enforce statically that `N` is non-zero with const generics
+        data.try_into().unwrap_or_else(|(_, err)| panic!("{err}"))
```
ralexstokes, line 127: I find `expect` to be a bit more literate for users, esp when we expect/know this won't panic

```suggestion
        data.try_into().map_err(|(_, err)| err).expect("any Vector can be constructed with default data")
```

1. we need the `map_err` bc I elect to return the "unparsable" data to the caller via the error channel, so we just need to drop the input data (I don't want to require `Debug` on it)
2. then we can just use `expect`

ralexstokes (approved):
this is excellent!

PR #57, opened 2023-05-17
Note possible checked_sub cases

doubledup:
For #22, I've looked through the places we've used subtraction and haven't found any cases where `checked_sub` is strictly necessary. The constraints for each case are documented in comments before the subtraction. I've also added a few constraints where necessary.

~Opening as a draft since I'm not sure whether we want to use checked_sub anyway, just in case the constraints change and we miss something. There is a small performance cost to checked_sub and I'm fairly certain it's unnecessary, so it's down to a cautiousness vs performance tradeoff.~


ssz-rs-derive/src/lib.rs
```diff
@@ -228,6 +229,7 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
                         let (_, end) = span[1];
 
                         container.__ssz_rs_set_by_index(index, &encoding[start..end])?;
+                        // checked_sub is unnecessary, as offsets are increasing
```
vgeddes, line 232: Small nit: 
```suggestion
                        // checked subtraction is unnecessary, as offsets are increasing; qed
```
I would apply the same style to the other changed lines.

ssz-rs/src/bitlist.rs
```diff
@@ -159,14 +161,22 @@ impl<const N: usize> Deserialize for Bitlist<N> {
         }
 
         let (last_byte, prefix) = encoding.split_last().unwrap();
+        if *last_byte == 0u8 {
```
ralexstokes, line 164: let's add a test for this case explicitly, as a trade off b/t not using `checked_sub` here but also not losing this invariant if someone changed this code

ssz-rs/src/bitlist.rs
```diff
@@ -159,14 +161,22 @@ impl<const N: usize> Deserialize for Bitlist<N> {
...
+            return Err(DeserializeError::InvalidByte(*last_byte));
+        }
+
         let mut result = BitlistInner::from_slice(prefix);
         let last = BitlistInner::from_element(*last_byte);
+        // checked subtraction is unnecessary, as last_byte != 0, so last.trailing_zeros <= 7; qed
```
ralexstokes, line 170: another thing we could do here is to make it clear these comments are for _safety_

```suggestion
        // SAFETY: checked subtraction is unnecessary, as last_byte != 0, so last.trailing_zeros <= 7; qed
```

ssz-rs/src/bitlist.rs
```diff
@@ -159,14 +161,22 @@ impl<const N: usize> Deserialize for Bitlist<N> {
...
+        // high_bit_index >= 1
```
ralexstokes, line 171:
```suggestion
        // therefore: high_bit_index >= 1
```

ssz-rs/src/merkleization/mod.rs
```diff
@@ -46,9 +46,10 @@ impl Display for MerkleizationError {
 impl std::error::Error for MerkleizationError {}
 
 pub fn pack_bytes(buffer: &mut Vec<u8>) {
-    let data_len = buffer.len();
-    if data_len % BYTES_PER_CHUNK != 0 {
-        let bytes_to_pad = BYTES_PER_CHUNK - data_len % BYTES_PER_CHUNK;
+    let incomplete_chunk_len = buffer.len() % BYTES_PER_CHUNK;
+    if incomplete_chunk_len != 0 {
+        // checked subtraction is unnecessary, as BYTES_PER_CHUNK > incomplete_chunk_len; qed
+        let bytes_to_pad = BYTES_PER_CHUNK - incomplete_chunk_len;
```
ralexstokes, line 52: this is better, thank you :)

ssz-rs/src/merkleization/mod.rs
```diff
@@ -265,27 +283,36 @@ mod tests {
...
         let node_count = 2 * leaf_count - 1;
+        // checked subtraction is unnecessary, as node_count >= leaf_count; qed
         let interior_count = node_count - leaf_count;
         let leaf_start = interior_count * BYTES_PER_CHUNK;
 
         let mut hasher = Sha256::new();
         let mut buffer = vec![0u8; node_count * BYTES_PER_CHUNK];
         buffer[leaf_start..leaf_start + chunks.len()].copy_from_slice(chunks);
-        let zero_chunk = [0u8; 32];
+        let zero_chunk = [0u8; BYTES_PER_CHUNK];
```
ralexstokes, line 295: thanks

ralexstokes (commented):
I need to make another pass on the merkle code but this is great!

ralexstokes:
> so it's down to a cautiousness vs performance tradeoff.

I think we are fine to just add comments like you have done; I would consider calling them out explicitly with a `SAFETY:` marker in the comments

ssz-rs-derive/src/lib.rs
```diff
@@ -192,8 +192,9 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
                     Some(field_name) => quote_spanned! { f.span() =>
                         let bytes_read = if <#field_type>::is_variable_size() {
                             let end = start + #BYTES_PER_LENGTH_OFFSET;
-                            let next_offset = u32::deserialize(&encoding[start..end])?;
-                            offsets.push((#i, next_offset as usize));
+                            let next_offset = u32::deserialize(&encoding[start..end])? as usize;
+                            offsets.last().map(|(_, previous_offset)| assert!(next_offset >= *previous_offset));
```
ralexstokes, line 196: this is a good point -- I think instead we should return an error here instead of panicking via the `assert!`

e.g. from the deserialization spec: https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#deserialization

> Offsets: out of order, out of range, mismatching minimum element size.

let's either patch in this PR or ill make an issue to fix before v1

ssz-rs/src/merkleization/mod.rs
```diff
@@ -129,10 +132,17 @@ fn merkleize_chunks_with_virtual_padding(
...
+                    // focus.len() = (i + 2 - parent_index) * BYTES_PER_CHUNK
+                    // and
+                    // i >= parent_index
+                    // so focus.len() >= 2 * BYTES_PER_CHUNK; qed
                     let children_index = focus.len() - 2 * BYTES_PER_CHUNK;
                     let (parent, children) = focus.split_at_mut(children_index);
                     let (left, right) = children.split_at_mut(BYTES_PER_CHUNK);
                     if parent.is_empty() {
+                        // TODO: Why not i == 0? Is there another case where the left node is used
+                        // to store the parent?
```
ralexstokes, line 145: I'd have to dig back into this code to answer this question; let's make an issue

ssz-rs/src/merkleization/mod.rs
```diff
@@ -145,9 +155,17 @@ fn merkleize_chunks_with_virtual_padding(
...
                         &mut layer[parent_index * BYTES_PER_CHUNK..(i + 1) * BYTES_PER_CHUNK];
+                    // checked subtraction is unnecessary:
+                    // focus.len() = (i + 1 - parent_index) * BYTES_PER_CHUNK
+                    // and
+                    // i >= parent_index
+                    // so focus.len() >= BYTES_PER_CHUNK; qed
                     let children_index = focus.len() - BYTES_PER_CHUNK;
-                    let (parent, children) = focus.split_at_mut(children_index);
-                    let (left, _) = children.split_at_mut(BYTES_PER_CHUNK);
+                    let (parent, left) = focus.split_at_mut(children_index);
```
ralexstokes, line 164: this will fail if `left` happens to be more than _just_ 32 bytes... let's verify before making this simplification

ssz-rs/src/merkleization/mod.rs
```diff
@@ -265,27 +283,36 @@ mod tests {
...
+            //             = ((2*i + 2 - i + 1) / 2) * BYTES_PER_CHUNK
+            //             = ((i + 3) / 2) * BYTES_PER_CHUNK
+            // and
+            // i >= 1
+            // so focus.len() >= 2 * BYTES_PER_CHUNK; qed
             let children_index = focus.len() - 2 * BYTES_PER_CHUNK;
             let (parent, children) = focus.split_at_mut(children_index);
-            let left = &children[0..BYTES_PER_CHUNK];
-            let right = &children[BYTES_PER_CHUNK..2 * BYTES_PER_CHUNK];
+            let (left, right) = children.split_at(BYTES_PER_CHUNK);
```
ralexstokes, line 315: let's verify that `children.len() == 32` _always_ and then I agree this is better

doubledup: We should have `children.len() == 64` (`2 * BYTES_PER_CHUNK`) always because:
- Above `(parent, children) = focus.split_at_mut(children_index) == focus.split_at_mut(focus.len() - 2 * BYTES_PER_CHUNK)`, so `children.len() <= 2 * BYTES_PER_CHUNK`.
and
- We know that `focus.len() >= 2 * BYTES_PER_CHUNK`, so we must have `children.len() >= 2 * BYTES_PER_CHUNK`.

ssz-rs-derive/src/lib.rs
doubledup: Oh wow, I added an `assert!` 🙈 Sorry, definitely better to return an error here.

ssz-rs/src/merkleization/mod.rs
doubledup: #63

ssz-rs/src/bitlist.rs
doubledup: Added a case to the [`decode_bitlist` test](https://github.com/ralexstokes/ssz-rs/pull/57/files#diff-7424acc2e06bfc6ea7bd3686c0618f6541553fca52fb1c71149a08ad0e93c52bR288)

ssz-rs-derive/src/lib.rs
```diff
@@ -192,14 +192,24 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
...
                             let target = encoding.get(start..end).ok_or_else(||
                                 ssz_rs::DeserializeError::ExpectedFurtherInput {
                                     provided: encoding.len() - start,
                                     expected: #BYTES_PER_LENGTH_OFFSET,
                                 }
                             )?;
-                            let next_offset = u32::deserialize(target)?;
-                            offsets.push((#i, next_offset as usize));
+                            let next_offset = u32::deserialize(target)? as usize;
+                            offsets.push((#i, next_offset));
```
ralexstokes, line 203: this `push` call should be after checking the offsets' increasing-ness right?

it looks like what we have now puts `next_offset == previous_offset` always

ssz-rs/src/bitlist.rs
```diff
@@ -279,6 +291,11 @@ mod tests {
         )
         .unwrap();
         assert_eq!(result, expected);
+
+        let bytes = vec![24u8, 0u8];
+        let result = Bitlist::<COUNT>::deserialize(&bytes).expect_err("test data is incorrect");
+        let expected = DeserializeError::InvalidByte(0u8);
+        assert_eq!(result.to_string(), expected.to_string());
```
ralexstokes, line 298: nice!

ralexstokes (commented):
had a question about offset checking for container deserialization

and still need to look at the merkle code

but everything else is looking great!

ssz-rs-derive/src/lib.rs
doubledup: Good catch!

ssz-rs-derive/src/lib.rs
```diff
@@ -193,14 +193,25 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
...
+                            let next_offset = u32::deserialize(target)? as usize;
+
+                            if let Some((_, previous_offset)) = offsets.last() {
+                                if next_offset < *previous_offset {
+                                    return Err(DeserializeError::OffsetNotIncreasing {
+                                        start: *previous_offset,
+                                        end: next_offset,
+                                    });
+                                }
+                            };
```
ralexstokes, line 212: style pref
```suggestion
                                    })
                                }
                            }
```

ssz-rs/src/merkleization/mod.rs
ralexstokes: I recall now that I *think* this is here because it just makes the math simpler if we ignore the node at `i==0`

I will re-verify but basically we are just saying "its ok to waste these 32 bytes" bc the code is much easier to implement

I don't know if a more sophisticated implementation will ever be needed but you can go beyond the relatively simple Merkle code here so I didn't feel like it was that important to put that much effort into this

a benchmark and a use case is all it would take to justify improving this however :)

ralexstokes: if this is in fact the case ill add a clarifying comment somewhere about the data structure; if the issue is just an offset by 1 then I should also be able to refactor very easily and then skip the "wasted" bytes

ssz-rs/src/merkleization/mod.rs
```diff
@@ -44,9 +44,11 @@ impl Display for MerkleizationError {
...
-    let data_len = buffer.len();
-    if data_len % BYTES_PER_CHUNK != 0 {
-        let bytes_to_pad = BYTES_PER_CHUNK - data_len % BYTES_PER_CHUNK;
+    let incomplete_chunk_len = buffer.len() % BYTES_PER_CHUNK;
+    if incomplete_chunk_len != 0 {
+        // SAFETY: checked subtraction is unnecessary,
+        // as BYTES_PER_CHUNK > incomplete_chunk_len; qed
+        let bytes_to_pad = BYTES_PER_CHUNK - incomplete_chunk_len;
         let pad = vec![0u8; bytes_to_pad];
         buffer.extend_from_slice(&pad);
```
ralexstokes, line 53: we can skip the allocation (via `vec!`) here

```suggestion
        buffer.resize(buffer.len() + bytes_to_pad, 0);
```

ssz-rs/src/merkleization/mod.rs
```diff
@@ -44,11 +44,12 @@ impl Display for MerkleizationError {
 impl std::error::Error for MerkleizationError {}
 
 pub fn pack_bytes(buffer: &mut Vec<u8>) {
```
ralexstokes, line 46:
```suggestion
// Ensures `buffer` can be exactly broken up into `BYTES_PER_CHUNK` chunks of bytes
// via padding any partial chunks at the end of `buffer`
pub fn pack_bytes(buffer: &mut Vec<u8>) {
```

ssz-rs/src/merkleization/mod.rs
ralexstokes: yep I see, thanks!

ssz-rs/src/merkleization/mod.rs
```diff
@@ -288,27 +309,38 @@ mod tests {
...
         let mut hasher = Sha256::new();
         let mut buffer = vec![0u8; node_count * BYTES_PER_CHUNK];
         buffer[leaf_start..leaf_start + chunks.len()].copy_from_slice(chunks);
-        let zero_chunk = [0u8; 32];
+        let zero_chunk = [0u8; BYTES_PER_CHUNK];
         for i in chunks.len()..leaf_count {
             let start = leaf_start + (i * BYTES_PER_CHUNK);
             let end = leaf_start + (i + 1) * BYTES_PER_CHUNK;
             buffer[start..end].copy_from_slice(&zero_chunk);
         }
```
ralexstokes: I don't think the bounds on this iterator are correct `chunks.len()..leaf_count` and it was fine only bc we are just writing zeros to memory that was already zero (initialized in `buffer`)

I deleted it anyway bc it didn't seem to be doing anything

ssz-rs/src/merkleization/mod.rs
ralexstokes: scratch all of this... 

ill move to the issue

PR #58, opened 2023-05-25
Make `no_std` build work with `serde` feature enabled

aeryz:
when `serde` feature is enabled in a `no_std` build, it fails due to the references to `std`. This PR makes `serde` feature compatible with `no_std`.

Note that `thiserror` is completely removed because it doesn't support `no_std`.

### DIFF
--- a/ssz-rs/src/list.rs
+++ b/ssz-rs/src/list.rs
@@ -10,8 +10,6 @@ use crate::{
 };
 #[cfg(feature = "serde")]
 use serde::ser::SerializeSeq;
-#[cfg(feature = "serde")]
-use std::marker::PhantomData;
 
 /// A homogenous collection of a variable number of values.
 #[derive(Clone)]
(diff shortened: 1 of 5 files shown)


ssz-rs/src/list.rs
```diff
@@ -12,7 +12,7 @@ use crate::{
 #[cfg(feature = "serde")]
 use serde::ser::SerializeSeq;
 #[cfg(feature = "serde")]
-use std::marker::PhantomData;
+use core::marker::PhantomData;
```
ralexstokes, line 15: can we follow the other types here and move `PhantomData` to the `crate::lib` and export one or the other depending on std or not?

ssz-rs/src/container.rs
```diff
@@ -1,6 +1,6 @@
 #[cfg(test)]
 mod tests {
-    use crate::prelude::*;
+    use crate::{lib::*, prelude::*};
```
ralexstokes, line 3: why do we need to pull in the `lib` imports into the tests now?

ralexstokes (commented):
a few questions but generally I'm open to this

ssz-rs/src/container.rs
aeryz: oh actually, we don't necessarily need to run the tests in no_std, so we don't need this as well. removing

ralexstokes (approved):
hey sorry for the delay here, there is an audit underway and I was trying to minimize how much `main` differed from the audit target but I'm losing that battle

this looks good!

PR #60, opened 2023-06-01
Replace index with get

doubledup:
This helps address #22, though I've done my best to avoid overlapping with #54 (I've left the case that's handled there in `ssz-rs-derive`).


ssz-rs/build.rs
```diff
@@ -24,7 +24,7 @@ fn compute_zero_hashes() -> [u8; MAX_MERKLE_TREE_DEPTH * BYTES_PER_CHUNK] {
 }
 
 // This function derives a set of bytes corresponding to "zero hashes" at build-time
-// in lieu of needing to declar any sort of runtime static memory or similar technique.
+// in lieu of needing to declare any sort of runtime static memory or similar technique.
```
ralexstokes, line 27: ty!

ssz-rs/src/merkleization/node.rs
```diff
@@ -1,9 +1,14 @@
 use crate::{lib::*, prelude::*, utils::write_bytes_to_lower_hex};
 
+use super::BYTES_PER_CHUNK;
+
+// TODO: Should we change Node's length to BYTES_PER_CHUNK?
+// They need to match for merkleization in ssz-rs-derive's hash_tree_root to work.
```
ralexstokes, line 6: yes, this was an oversight, perhaps I defined `Node` before I had `BYTES_PER_CHUNK` to rely on

ssz-rs/src/merkleization/mod.rs
```diff
@@ -112,11 +116,15 @@ fn merkleize_chunks_with_virtual_padding(
...
     debug_assert!(leaf_count.next_power_of_two() == leaf_count);
+    // this holds as long as usize is no longer than u64
+    debug_assert!((leaf_count.trailing_zeros() as usize) < MAX_MERKLE_TREE_DEPTH);
 
     let height = leaf_count.trailing_zeros() + 1;
 
     if chunk_count == 0 {
         let depth = height - 1;
+        // index is safe as long as depth == leaf_count.trailing_zeros() < MAX_MERKLE_TREE_DEPTH;
+        // qed
```
ralexstokes, line 127: I like this more compact style you had
```suggestion
        // index is safe as long as depth == leaf_count.trailing_zeros() < MAX_MERKLE_TREE_DEPTH; qed
```

ssz-rs/src/list.rs
```diff
@@ -240,6 +242,7 @@ where
             for (i, elem) in self.data.iter_mut().enumerate() {
                 let chunk = elem.hash_tree_root()?;
                 let range = i * BYTES_PER_CHUNK..(i + 1) * BYTES_PER_CHUNK;
+                // index is safe as long as Node::len() == BYTES_PER_CHUNK
```
ralexstokes, line 245: the `Node` type will always be `BYTES_PER_CHUNK` in size, so we can drop this comment, esp if you feel like that invariant is safely enforced elsewhere

ssz-rs-derive/src/lib.rs
```diff
@@ -170,6 +170,7 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
                 // "tuple" struct
                 // only support the case with one unnamed field, to support "newtype" pattern
                 Fields::Unnamed(ref fields) => {
+                    // index is safe because Punctuated always has a first element; qed
```
ralexstokes, line 173: fwiw im less concerned about panics in the proc macro itself

ssz-rs/src/bitlist.rs
```diff
@@ -163,6 +163,9 @@ impl<const N: usize> Deserialize for Bitlist<N> {
         let last = BitlistInner::from_element(*last_byte);
         let high_bit_index = 8 - last.trailing_zeros();
 
+        // index should be safe because the last_byte shouldn't be 0
+        // for a check for this condition, see
+        // https://github.com/ralexstokes/ssz-rs/pull/57/files#diff-7424acc2e06bfc6ea7bd3686c0618f6541553fca52fb1c71149a08ad0e93c52bR164-R175
```
ralexstokes, line 168: lets merge this in w/ #57 and remove the link to github

ssz-rs-derive/src/lib.rs
```diff
@@ -224,10 +225,17 @@ fn derive_deserialize_impl(data: &Data) -> TokenStream {
                     offsets.push((dummy_index, encoding.len()));
 
                     for span in offsets.windows(2) {
+                        // indexes are safe because span is a pair; qed
                         let (index, start) = span[0];
                         let (_, end) = span[1];
 
-                        container.__ssz_rs_set_by_index(index, &encoding[start..end])?;
+                        let target = encoding.get(start..end).ok_or(
```
ralexstokes, line 232: did this in #54 just to avoid the extra computation unless needed
```suggestion
                        let target = encoding.get(start..end).ok_or_else(||
```

ssz-rs-derive/src/lib.rs
```diff
@@ -369,12 +380,22 @@ fn derive_merkleization_impl(data: &Data) -> TokenStream {
...
                 None => quote_spanned! { f.span() =>
                     let chunk = self.0.hash_tree_root()?;
                     let range = #i*#BYTES_PER_CHUNK..(#i+1)*#BYTES_PER_CHUNK;
-                    chunks[range].copy_from_slice(chunk.as_ref());
+                    let chunks_byte_count = chunks.len();
+                    // index is safe as long as Node::len() == BYTES_PER_CHUNK
+                    chunks
+                        .get_mut(range)
+                        .ok_or(MerkleizationError::InputInsufficient(chunks_byte_count))?
+                        .copy_from_slice(chunk.as_ref());
```
ralexstokes, line 398: somewhat strong preference for leaving these alone

very open to either a comment `// SAFETY: ...` or some kind of `debug_assert!` that the invariant holds

but this was the intention (impl is auto-generated and can just directly read/write the buffer)

ssz-rs/src/array.rs
```diff
@@ -80,6 +80,7 @@ macro_rules! define_ssz_for_array_of_size {
                     for (i, elem) in self.iter_mut().enumerate() {
                         let chunk = elem.hash_tree_root()?;
                         let range = i * BYTES_PER_CHUNK..(i + 1) * BYTES_PER_CHUNK;
+                        // index is safe as long as Node::len() == BYTES_PER_CHUNK
```
ralexstokes, line 83: I think this assumption is strong enough to not warrant a comment but TBH this helps anyone who doesn't know the SSZ spec that well so I don't mind if we leave it in

ssz-rs/src/merkleization/mod.rs
```diff
@@ -23,6 +23,7 @@ pub trait Merkleized {
 pub enum MerkleizationError {
     SerializationError(SerializeError),
     InputExceedsLimit(usize),
+    InputInsufficient(usize),
```
ralexstokes, line 26: drop this if we drop it everywhere else

ssz-rs/src/de.rs
```diff
@@ -48,6 +55,8 @@ impl Display for DeserializeError {
             ),
             DeserializeError::InvalidInstance(err) => write!(f, "invalid instance: {err}"),
             DeserializeError::InvalidType(err) => write!(f, "invalid type: {err}"),
+            DeserializeError::IncompleteLengthOffsets(err) => write!(f, "incomplete length offsets: {err}"),
+            DeserializeError::OffsetNotIncreasing { start, end } => write!(f, "empty span from {start} to {end} bytes"),
```
ralexstokes, line 59: I think this is a little clearer
```suggestion
            DeserializeError::OffsetNotIncreasing { start, end } => write!(f, "offset points to {end} bytes before {start} bytes"),
```

ssz-rs/src/de.rs
```diff
@@ -93,26 +102,45 @@ where
...
+        return Err(DeserializeError::IncompleteLengthOffsets(offsets_len));
+    }
 
-    let offsets = &mut encoding[..data_pointer]
+    let offsets = &mut encoding
+        .get(..offsets_len)
+        .ok_or(DeserializeError::ExpectedFurtherInput {
+            provided: encoding.len(),
+            expected: offsets_len,
+        })?
```
ralexstokes, line 127: we can drop this as we just verified `encoding.len() >= offsets_len` on line 112

ssz-rs/src/de.rs
```diff
@@ -93,26 +102,45 @@ where
...
+    if encoding.len() < offsets_len {
         return Err(DeserializeError::ExpectedFurtherInput {
             provided: encoding.len(),
-            expected: data_pointer,
+            expected: offsets_len,
         })
     }
+    if offsets_len % BYTES_PER_LENGTH_OFFSET != 0 {
+        return Err(DeserializeError::IncompleteLengthOffsets(offsets_len));
+    }
```
ralexstokes, line 120: good catch!

ssz-rs/src/de.rs
```diff
@@ -48,6 +55,8 @@ impl Display for DeserializeError {
             ),
             DeserializeError::InvalidInstance(err) => write!(f, "invalid instance: {err}"),
             DeserializeError::InvalidType(err) => write!(f, "invalid type: {err}"),
+            DeserializeError::IncompleteLengthOffsets(err) => write!(f, "incomplete length offsets: {err}"),
```
ralexstokes, line 58:
```suggestion
            DeserializeError::InvalidOffsetsLength(len) => write!(f, "the offsets length provided {len} is not a multiple of the size per length offset {BYTES_PER_LENGTH_OFFSET} bytes"),
```

ssz-rs/src/de.rs
```diff
@@ -93,26 +102,45 @@ where
...
+    }
 
-    let offsets = &mut encoding[..data_pointer]
+    let offsets = &mut encoding
+        .get(..offsets_len)
+        .ok_or(DeserializeError::ExpectedFurtherInput {
+            provided: encoding.len(),
+            expected: offsets_len,
+        })?
         .chunks_exact(BYTES_PER_LENGTH_OFFSET)
```
ralexstokes, line 128: noting we don't need to check the `chunks_exact` iterator (there is a `remainder` method I just discovered) as we verified the sizes align above

ssz-rs/src/list.rs
```diff
@@ -150,6 +150,7 @@ where
     type Output = <Idx as SliceIndex<[T]>>::Output;
 
     fn index(&self, index: Idx) -> &Self::Output {
+        // Index::index implementations may panic when the given index is out of bounds
```
ralexstokes, line 153: I would lean towards leaving these comments out as the idea is to also satisfy the `Index::index` interface here but if you want them in, we can keep them

ditto for the rest of these that just lean on the wrapping interface for invariants

ssz-rs/src/ser.rs
```diff
@@ -74,6 +86,7 @@ pub fn serialize_composite_from_components(
             total_bytes_written += part.len();
             buffer.append(part);
         } else {
+            // index is safe because variable_lengths.len() >= fixed.len()
             let variable_lengths_sum = variable_lengths[0..i].iter().sum::<usize>();
             let length = (fixed_lengths_sum + variable_lengths_sum) as u32;
             let mut offset_buffer = Vec::with_capacity(4);
```
ralexstokes, line 92:
```suggestion
            let mut offset_buffer = Vec::with_capacity(BYTES_PER_LENGTH_OFFSET);
```

ralexstokes: and the same thing on `total_bytes_written` -- I can fix this if you want

ssz-rs/src/ser.rs
```diff
@@ -66,6 +71,13 @@ pub fn serialize_composite_from_components(
     if total_size as u64 >= MAXIMUM_LENGTH {
         return Err(SerializeError::MaximumEncodedLengthExceeded(total_size))
     }
+    // TODO: should this be == instead?
+    if fixed.len() > variable_lengths.len() {
```
ralexstokes, line 75: as currently written, `fixed` and `variable_lengths` are _always_ equal in size, it is an error via the caller if this is not the case... could do w/ a `debug_assert!` possibly although after looking at this code I may try to simplify some stuff...

ssz-rs/src/ser.rs
```diff
@@ -74,6 +86,7 @@ pub fn serialize_composite_from_components(
             total_bytes_written += part.len();
             buffer.append(part);
         } else {
+            // index is safe because variable_lengths.len() >= fixed.len()
```
ralexstokes, line 89: safe bc `variable_lengths.len() == fixed.len()` always

ralexstokes (commented):
this is amazing! left some comments/questions

ssz-rs/src/ser.rs
ralexstokes: see https://github.com/ralexstokes/ssz-rs/pull/65

ssz-rs/src/merkleization/node.rs
```diff
@@ -1,9 +1,14 @@
 use crate::{lib::*, prelude::*, utils::write_bytes_to_lower_hex};
 
+use super::BYTES_PER_CHUNK;
```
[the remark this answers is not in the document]
doubledup: Definitely, I think this was an auto-import. I'll merge it with the `crate` import above.

ssz-rs/src/merkleization/mod.rs
doubledup: This was wrapped because of the [max comment width](https://github.com/ralexstokes/ssz-rs/blob/main/rustfmt.toml#L4). We can rephrase it though:
```suggestion
        // index is safe while depth == leaf_count.trailing_zeros() < MAX_MERKLE_TREE_DEPTH; qed
```

ssz-rs/src/list.rs
doubledup: Definitely happy to drop it with `Node`'s length defined as `BYTES_PER_CHUNK`

ssz-rs/src/array.rs
doubledup: I'm happy to remove it with `Node`'s length being defined as `BYTES_PER_CHUNK`

ssz-rs/src/de.rs
doubledup: Isn't `chunks_exact` here so that we feed 4-byte slices into the `map` below? Else `map` would operate on each byte.

ssz-rs/src/ser.rs
doubledup: Switched to `debug_assert_eq!` here, so it'll be easier to merge with #65

ralexstokes:
`cargo` is being opinionated about the whitespace formatting of the comments... ill go fix the lint

ssz-rs/src/de.rs
ralexstokes: yep! I was referring to making sure the input is aligned w/ `BYTES_PER_LENGTH_OFFSET` e.g. `input.len() % BYTES_PER_LENGTH_OFFSET == 0`

ssz-rs/src/de.rs
```diff
@@ -93,26 +102,41 @@ where
...
     let mut result = Vec::with_capacity(element_count);
     for span in offsets.windows(2) {
+        // SAFETY: index is safe because span is a pair; qed
         let start = span[0];
         let end = span[1];
+        if start > end {
+            return Err(DeserializeError::OffsetNotIncreasing { start, end })
+        }
+
+        // SAFETY: index is safe because start <= end; qed
```
ralexstokes, line 139: and `start` can be at *least* 0 and `end` can be at *most* the length of the encoding

ssz-rs/src/de.rs
```diff
@@ -93,26 +102,41 @@ where
...
+    let offsets = &mut encoding[..offsets_len]
         .chunks_exact(BYTES_PER_LENGTH_OFFSET)
         .map(|chunk| u32::deserialize(chunk).map(|offset| offset as usize))
         .collect::<Result<Vec<usize>, DeserializeError>>()?;
     offsets.push(encoding.len());
 
-    let element_count = data_pointer / BYTES_PER_LENGTH_OFFSET;
+    let element_count = offsets_len / BYTES_PER_LENGTH_OFFSET;
     let mut result = Vec::with_capacity(element_count);
     for span in offsets.windows(2) {
```
ralexstokes, line 131: also worth noting that I believe as far as this code can be reached, there will also be at least 2 elements in `offsets` so it is fine to call `windows(2)` here and not worry about skipping/missing some data

ralexstokes (approved):
excellent updates, thanks!

ssz-rs/src/de.rs
doubledup: There's technically a case with 1 element in `offsets`, but it doesn't look like a problem:

If `offsets_len == 0` then `offsets == vec![encoding.len()]`. In this case we return an empty `Vec` (both `len == 0` and `cap == 0`). This seems reasonable though, because this function is for variable sized composite types like `List` that are allowed to be empty.

PR #75, opened 2023-06-09
Ignore ssz-rs-derive & ssz-rs-test-gen for coverage

doubledup:
Looks like coverage is [over 70% for `ssz-rs`](https://app.codecov.io/gh/ralexstokes/ssz-rs) (see the per-directory breakdown at the bottom of the page), so ignoring these directories will give us a clearer picture of coverage in PRs & the CodeCov UI.

@ralexstokes This will override the [global yaml](https://docs.codecov.com/docs/codecov-yaml#global-yaml), but only for the keys mentioned in `codecov.yml`.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,3 @@
+ignore:
+  - "ssz-rs-derive/**/*"
+  - "ssz-rs-test-gen/**/*"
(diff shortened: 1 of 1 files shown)


ralexstokes (approved):
excellent, thanks!

PR #79, opened 2023-06-20
Replaces debug_assert_eq with size check errors

claravanstaden:
From audit: 

> We recommend implementing mentioned checks with an error handling logic in order to
perform them also in production code.

### DIFF
--- a/ssz-rs/src/ser.rs
+++ b/ssz-rs/src/ser.rs
@@ -42,7 +42,9 @@ impl Display for SerializeError {
             ),
             SerializeError::InvalidInstance(err) => write!(f, "invalid instance: {err}"),
             SerializeError::InvalidType(err) => write!(f, "invalid type: {err}"),
-            SerializeError::UnexpectedSize(size1, size2) => write!(f, "unexpected size: {size1} not equal to {size2}"),
+            SerializeError::UnexpectedSize(size1, size2) => {
+                write!(f, "unexpected size: {size1} not equal to {size2}")
+            }
         }
     }
 }
--- a/ssz-rs/src/ser.rs
+++ b/ssz-rs/src/ser.rs
@@ -17,6 +17,8 @@ pub enum SerializeError {
     InvalidInstance(InstanceError),
     /// An invalid type was encountered.
     InvalidType(TypeError),
+    /// An unexpected size sanity check was encountered.
+    UnexpectedSize(usize, usize),
 }
 
 impl From<InstanceError> for SerializeError {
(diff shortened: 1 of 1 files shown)


ssz-rs/src/ser.rs
```diff
@@ -71,14 +78,15 @@ pub fn serialize_composite_from_components(
 
     // SAFETY: `fixed_lengths_sum` fits in `u32` if the total size check holds
     let mut running_length = fixed_lengths_sum as u32;
-    debug_assert_eq!(fixed.len(), variable_lengths.len());
```
claravanstaden, line 74: This seems to be a duplicate check of the check in line 65, and the variables don't change, so I removed this check.

ralexstokes: ah thank you! yeah this seems like copypasta slipped in

ralexstokes (commented):
hmm ok, sorry about this; it looks we can clean up this code in a few ways:

we can add the error type and turn the debug assertions into actual errors.

however, my intent was that this is an invariant that should always hold and so it is a user error to violate it; thus, the rationale to have a check during development but not during production (as any possible errors should have already been caught)

so, when I think about how to remove this, while also increasing the safety of this function, I think we should actually just require something like a `parts: Vec<Part>` input instead of `fixed` and `variable_lengths`

with:
```rust
enum Part {
  Fixed(Vec<u8>),
  Variable(usize),
}
```

ssz-rs/src/ser.rs
```diff
@@ -71,14 +78,15 @@ pub fn serialize_composite_from_components(
...
     for (part_opt, variable_length) in fixed.iter_mut().zip(variable_lengths) {
         if let Some(part) = part_opt {
             buffer.append(part);
         } else {
             // SAFETY: `variable_length` fits in `u32` if the total size check holds
             let bytes_written = running_length.serialize(buffer)?;
-            debug_assert_eq!(bytes_written, BYTES_PER_LENGTH_OFFSET);
+            if bytes_written != BYTES_PER_LENGTH_OFFSET {
+                return Err(SerializeError::UnexpectedSize(bytes_written, BYTES_PER_LENGTH_OFFSET))
+            }
```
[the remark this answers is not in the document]
ralexstokes: right -- the only way to trigger this error is if the code above changes; otherwise we know statically that `bytes_written` will be `BYTES_PER_LENGTH_OFFSET` and hence why I went w/ the `debug_assert` here

I kind of lean towards just leaving this one part alone

PR #81, opened 2023-06-23
harden typing to increase safety of serialization routine

ralexstokes:
minor but also skips a redundant iteration over the variable lengths summation that can be done inline like the fixed lengths summation

### DIFF
--- a/ssz-rs/src/ser.rs
+++ b/ssz-rs/src/ser.rs
@@ -11,8 +11,8 @@ const MAXIMUM_LENGTH: u64 = 2u64.pow((8 * BYTES_PER_LENGTH_OFFSET) as u32);
 /// Serialization errors.
 #[derive(Debug)]
 pub enum SerializeError {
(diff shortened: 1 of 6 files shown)


ssz-rs/src/ser.rs
```diff
@@ -55,32 +55,39 @@ pub trait Serialize {
...
-
-            running_length += variable_length as u32;
+    for part in parts {
+        match part {
+            Part::Fixed(mut data) => {
+                buffer.append(&mut data);
+            }
+            Part::Offset(offset) => {
+                let bytes_written = running_length.serialize(buffer)?;
+                debug_assert_eq!(bytes_written, BYTES_PER_LENGTH_OFFSET);
```
ralexstokes, line 86: this kind of just moves the problem elsewhere but we could change the defn of `Part`:

```rust
enum Part {
  Fixed(Vec<u8>),
  Offset(u32),
}
```

PR #84, opened 2023-06-27
skip some duplicate checks when deserializing fixed types

ralexstokes:
from Oak Security audit:

> When deserializing Array, List, and Vector objects, depending on whether they are fixed or variable, one of two functions is called within deserialize_homogeneous_composite - deserialize_variable_homogeneous_composite or deserialize_fixed_homogeneous_composite.
The latter is called when the type is fixed, and the corresponding deserialize functions have already validated the length of the object to be a multiple of the default value of that type obtained using T::size_hint. An example for the Array type is the validation performed in ssz-rs/src/array.rs:51-65.
Nevertheless, the deserialize_fixed_homogeneous_composite function in lines 71-78 validates whether the modulo of the length of the deserialized object and the default size for its type is different from zero. Bearing in mind the fact that in the previous step this size was multiplied by N, it means that consequently there is no possibility that the modulo will be different from zero. Ultimately, it is therefore a redundant piece of code.

### DIFF
--- a/ssz-rs/src/list.rs
+++ b/ssz-rs/src/list.rs
@@ -188,6 +188,18 @@ where
     T: SimpleSerialize,
 {
     fn deserialize(encoding: &[u8]) -> Result<Self, DeserializeError> {
+        if !T::is_variable_size() {
+            let remainder = encoding.len() % T::size_hint();
+            if remainder != 0 {
+                return Err(DeserializeError::AdditionalInput {
+                    provided: encoding.len(),
+                    // SAFETY: checked subtraction is unnecessary, as encoding.len() > remainder;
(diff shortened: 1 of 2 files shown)


ssz-rs/src/list.rs
```diff
@@ -188,6 +188,18 @@ where
     T: SimpleSerialize,
 {
     fn deserialize(encoding: &[u8]) -> Result<Self, DeserializeError> {
+        if !T::is_variable_size() {
```
ralexstokes, line 191: note, this check is not entirely redundant but if we want to remove the duplicity for `Vector` and array types, then we can do something like this

PR #88, opened 2023-06-30
add `Indexed` trait to compute generalized indices

ralexstokes:
part I of adding multiproofs:

adds an `Indexed` trait so types can compute their own generalized indices from statically defined paths into each SSZ data structure

and pulls in earlier multiproof code from #2.

TODO:
- [ ] double-check implementation and things we support (e.g. no recursive custom types for now)
- [ ] port `Container` and `Union` implementations to proc derive macro


ralexstokes:
something to check out: https://github.com/lightclient/proof

ralexstokes:
closing in lieu of #120

PR #90, opened 2023-07-02
feat: derive Hash for merkle node

jacobkaufmann:
while `Node` represents a hash itself, it would be useful for `Node` to implement `Hash`. for example, there are multiple instances in the ethereum consensus spec where we would like to key by the [`Root`](https://github.com/ralexstokes/ethereum-consensus/blob/46ca1716f2b1651e59b8a565b5bbcadf6b052fa5/src/primitives.rs#L8) of some data structure.


ralexstokes (approved):
makes sense, thank you!

PR #92, opened 2023-07-09
Derive serialize and deserialize

zsluedem:
1. Adding derive only for ~~Serialize and Deserialize~~ Serializable
~~2. relaxing the type constraint on List<T, N> and Vec<T, N> so that pure Serialize and Deserialize could be used directly too.~~


ssz-rs/src/de.rs
```diff
@@ -66,7 +65,7 @@ pub trait Deserialize {
 
 fn deserialize_fixed_homogeneous_composite<T>(encoding: &[u8]) -> Result<Vec<T>, DeserializeError>
 where
-    T: SimpleSerialize,
+    T: Deserialize + crate::Sized,
```
ralexstokes, line 68: minor, but let's just import `Sized` at the top and drop the ref to `crate` here

ralexstokes (commented):
nice work!

the macros look good; it would be nice if we can make them a bit DRYer somehow but I don't see a great way to do that without making the code a bit worse somewhere else

I don't love that we are dropping the trait bounds on `List<T>` and `Vector<T>`... the idea is to maintain type safety by only allowing someone to make a `List` or `Vector` with some other type that also implements `SimpleSerialize` -- the types don't really make sense without this

one option here is to drop the Merkle stuff and only require the serialization traits

```rust
trait Serializable: Serialize + Deserialize + Sized {}
```

and require `Serializable`

ralexstokes:
do you have an example use case where you would only want the serde functionality and the merkle stuff is a strict problem?

zsluedem:
> do you have an example use case where you would only want the serde functionality and the merkle stuff is a strict problem?

Hm... I am working on the p2p part of the rust bundler of erc4337. The p2p spec requires ssz encoding for [these types](https://github.com/eth-infinitism/bundler-spec/blob/main/p2p-specs/p2p-interface.md#container-specifications) and the merkle stuff is not needed here currently.

I implemented the encoding part for UserOp and didn't implement for merkle function but I still want to derive the serialize function only for more types.

ssz-rs/src/de.rs
zsluedem: It is conflicting the builtin `Sized` here https://github.com/ralexstokes/ssz-rs/blob/main/ssz-rs/src/de.rs#L64. Nvm, I renamed it.

ssz-rs-derive/src/lib.rs
```diff
@@ -517,6 +517,69 @@ fn validate_derive_data(data: &Data) {
...
+    let data = &input.data;
+    validate_derive_data(data);
+
+    let name = &input.ident;
+    let generics = &input.generics;
+
+    let serialize_impl = derive_serialize_impl(data);
+
+    let expansion =
+        impl_based_on_generics(generics, name, serialize_impl, quote! {ssz_rs::Serialize});
```
zsluedem, line 556: @ralexstokes Make some efforts to make it dryer.

ssz-rs/src/de.rs
```diff
@@ -2,7 +2,7 @@ use crate::{
     error::{InstanceError, TypeError},
     lib::*,
     ser::BYTES_PER_LENGTH_OFFSET,
-    SimpleSerialize,
+    Sized as SSZSized,
```
ralexstokes, line 5: I think we can just import `Sized` here

if you think that is confusing w/ the `std::Sized` then I have a slight preference for using `crate::Sized` down below instead of this `SSZSized` symbol

ralexstokes (commented):
hey! this is a great start

but while reviewing what we can do here I see a few places to clean up the trait hierarchy

I do see the value in splitting off the Merkleized trait, so give me a few days to land this on `main`; thanks for being patient as well :)

ralexstokes:
@zsluedem did some work on `main` to simplify the type constraints and also split off the serialization bit from the merkle bit

would love to see a PR for your updates to the proc macro here (may be easier to just cherry-pick to a new PR)

ssz-rs-derive/src/lib.rs
```diff
@@ -392,44 +395,27 @@ fn validate_derive_data(data: &Data) {
...
     let size_hint_impl = derive_size_hint_impl(data);
-    let merkleization_impl = derive_merkleization_impl(data);
 
-    let impl_impl = if generics.params.is_empty() {
-        quote! { impl }
-    } else {
-        quote! { impl #generics }
-    };
-
-    let name_impl = if generics.params.is_empty() {
```
zsluedem, line 416: Actually, this condition would not be necessary.  `split_for_impl` would do the trick and leave blank if type generic is empty on `quote!`.

ssz-rs-derive/src/lib.rs
```diff
@@ -392,44 +395,27 @@ fn validate_derive_data(data: &Data) {
...
-#[proc_macro_derive(SimpleSerialize)]
-pub fn derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
-    let input = parse_macro_input!(input as DeriveInput);
-
-    let data = &input.data;
-    validate_derive_data(data);
-
-    let name = &input.ident;
-    let generics = &input.generics;
+fn impl_serilizable(
```
ralexstokes, line 398:
```suggestion
fn impl_serializable(
```

ssz-rs/examples/serializable.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(PartialEq, Eq, Debug, Default, Serializable)]
+struct ComplexTestStruct {
+    a: u16,
+    b: List<u16, 128>,
+    c: u8,
+    d: List<u8, 256>,
+    e: VarTestStruct,
+    f: Vector<FixedTestStruct, 4>,
+    g: Vector<VarTestStruct, 2>,
+}
```
ralexstokes, line 37: instead of duplicating all of this code, let's just add one example struct to `examples/another_container.rs` that just derives `Serializable`

even just something like

```rust
#[derive(PartialEq, Eq, Debug, Default, Serializable)]
struct ComplexTestStruct {
    a: u16,
    b: List<u16, 128>,
    c: u8,
    d: List<u8, 256>,
    e: VarTestStruct,
    f: Vector<FixedTestStruct, 4>,
    g: Vector<VarTestStruct, 2>,
    h: u32,
}
```

and then show ser/de in that example code

ralexstokes: can you also add this example as a test to `ssz_rs::container::tests` and demonstrate a roundtrip?

that will give us some assurance the macro is working correctly

ssz-rs-derive/src/lib.rs
```diff
@@ -438,12 +424,47 @@ pub fn derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+
+    let data = &input.data;
+    validate_derive_data(data);
+
+    let name = &input.ident;
+    let generics = &input.generics;
+
+    let (impl_generics, ty_generics, _) = generics.split_for_impl();
+    let expand = impl_serilizable(data, name, &impl_generics, &ty_generics);
+    proc_macro::TokenStream::from(expand)
```
ralexstokes, line 442: nitpick

```suggestion
    let expansion = impl_serilizable(data, name, &impl_generics, &ty_generics);
    proc_macro::TokenStream::from(expansion)
```

ralexstokes (changes requested):
great updates!

let's add a test, refine the example and I had a few minor nitpicks on naming

ralexstokes (changes requested):
this is looking great!

can you also add some docs here: https://github.com/ralexstokes/ssz-rs/blob/main/ssz-rs-derive/README.md

something like:

```
This package provides two proc derive macros: `SimpleSerialize` and `Serializable`.

`SimpleSerialize` derives the functionality required to implement the main package's `SimpleSerialize` trait.

`Serializable` only derives the encoding and decoding routines, in the event a user does not want to pull in the merkleization machinery.
```

ralexstokes (approved):
lgtm!

ssz-rs-derive/src/lib.rs
```diff
@@ -392,44 +395,27 @@ fn validate_derive_data(data: &Data) {
...
-#[proc_macro_derive(SimpleSerialize)]
-pub fn derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
-    let input = parse_macro_input!(input as DeriveInput);
-
-    let data = &input.data;
-    validate_derive_data(data);
-
-    let name = &input.ident;
-    let generics = &input.generics;
+fn impl_serializable(
```
ralexstokes, line 398:
```suggestion
fn derive_serializable_impl(
```

ssz-rs-derive/src/lib.rs
```diff
@@ -438,12 +424,47 @@ pub fn derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    let input = parse_macro_input!(input as DeriveInput);
+
+    let data = &input.data;
+    validate_derive_data(data);
+
+    let name = &input.ident;
+    let generics = &input.generics;
+
+    let (impl_generics, ty_generics, _) = generics.split_for_impl();
+    let expansion = impl_serializable(data, name, &impl_generics, &ty_generics);
```
ralexstokes, line 441:
```suggestion
    let expansion = derive_serializable_impl(data, name, &impl_generics, &ty_generics);
```

ssz-rs-derive/src/lib.rs
```diff
@@ -438,12 +424,47 @@ pub fn derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    let data = &input.data;
+    validate_derive_data(data);
+
+    let name = &input.ident;
+    let generics = &input.generics;
+    let merkleization_impl = derive_merkleization_impl(data);
+
+    let (impl_generics, ty_generics, _) = generics.split_for_impl();
+
+    let serializable_impl = impl_serializable(data, name, &impl_generics, &ty_generics);
```
ralexstokes, line 458:
```suggestion
    let serializable_impl = derive_serializable_impl(data, name, &impl_generics, &ty_generics);
```

ralexstokes (approved):
one naming nitpick

zsluedem:
Thanks

PR #104, opened 2023-09-18
feat: migrate to alloy primitives

### LINKED ISSUE #94: migrate to `alloy` primitives
https://github.com/alloy-rs/core/tree/main/crates/primitives

this is a bigger task but should pay dividends in the long run to unify foundations across the communities

PatStiles:
Seeks to close #94


PatStiles:
@ralexstokes Let me know if i am on the right track here.

PatStiles:
@ralexstokes let me know if I'm on the right track. I've successfully remvoed U256.

ralexstokes (changes requested):
we can use the built-in rust types

the only thing we need to do here is migrate the currently defined `U256` type to the one from alloy

e.g. we won't need to change any of the calling code, so all of the `use alloy_primitives::U256` should disappear

ssz-rs/src/lib.rs
```diff
@@ -127,6 +127,7 @@ pub mod prelude {
     // with a simple `prelude` import
     pub use crate as ssz_rs;
     pub use ssz_rs_derive::{Serializable, SimpleSerialize};
+    pub use alloy_primitives::U256;
```
ralexstokes, line 130: so this should re-export our type (which just happens to be alloy's `U256`

```suggestion
    pub use uint::U256;
```

(although this line should already be present above somehow)

ralexstokes:
wrapped this up in #113

PR #119, opened 2023-12-07
Bump alloy dependency

idatsy:
Bump version up to 0.5. Needed because the reth stack uses `^0.5` so updates to `mev-rs` leveraging `reth` crates are not possible due to versioning conflicts in `ethereum-consensus`. After this is merged the `ssz-rs` dependency in `ethereum-consesus` can be pumped to latest which should resolve the conflicts?


ralexstokes (approved):
makes sense! thank you

PR #127, opened 2024-02-26
Replaces `log_2` logic with `checked_ilog2`

### LINKED ISSUE #126: Replace `log_2` with std lib impl
https://github.com/ralexstokes/ssz-rs/blob/6329986d14e71a98a35cc9323f88cbef0332b6dd/ssz-rs/src/merkleization/multiproofs/generalized_index.rs#L12

should be able to use this instead: https://doc.rust-lang.org/std/primitive.usize.html#method.checked_ilog2

EchoAlice:
Fixes #126

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
   coverage:
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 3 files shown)


ralexstokes (changes requested):
looks good! just need to tidy `lib.rs` (see clippy output)

PR #145, opened 2024-04-01
migrate `Node` impl to `alloy-primitives` type and simplify

### LINKED ISSUE #114: explore migrating `merkleization::Node` to the `alloy-primitives` fixed bytes 32 type
it could be nice to base all of our primitives around `alloy` so things are uniform and this repo can minimize how much code it maintains

given that `Node` is just a wrapper around `[u8;32]` im not sure this will be worth it, but it is certainly worth investigating

ralexstokes:
Fixes #114.

Primarily, migrate `Node` to `alloy_primitives::B256` to drop how much custom code we need

Also starts to lay the ground work for a similar refactor in `ethereum-consensus` to drop more custom serde code there (particularly around bytes and hex munging).


ssz-rs/src/lib.rs
```diff
@@ -124,9 +120,30 @@ mod exports {
         },
         ser::{Serialize, SerializeError},
         uint::U256,
-        utils::{deserialize, serialize},
         vector::Vector,
     };
+
+    /// `serialize` is a convenience function for taking a value that
+    /// implements `SimpleSerialize` and attempting to encode it to
```
ralexstokes, line 127:
```suggestion
    /// implements `SimpleSerialize` and attempts to encode it to
```

PR #153, opened 2024-04-25
Add support for verifying compact multiproofs

rista404:
Hi! Thanks for maintaining this great library. We were using it in a (soon to be open-source) project but found ourselves missing the compact multiproof support. This PR adds the support for verifying such proofs.

You can read the spec here: https://github.com/ethereum/consensus-specs/pull/3148
You can also see a full TypeScript implementation here: https://github.com/ChainSafe/ssz/pull/292

I think those two PRs explain why compact proofs are useful! Lodestar, which we interact with, is already using them in production as the only format for certain proofs.

I've marked the implementation as experimental and added some very basic tests. I could do a follow-up PR that adds proof generation as well as more tests. Let me know if something needs changing!

### DIFF
--- a/ssz-rs/src/merkleization/compact_multiproofs.rs
+++ b/ssz-rs/src/merkleization/compact_multiproofs.rs
@@ -1,0 +1,188 @@
+//! Experimental support for compact multiproofs.
+use crate::{
+    lib::*,
+    merkleization::{
+        multiproofs::{get_helper_indices, get_path_indices},
+        GeneralizedIndex, MerkleizationError as Error, Node,
+    },
+};
+use sha2::{Digest, Sha256};
+
+fn compute_proof_indices(indices: &[GeneralizedIndex]) -> Vec<GeneralizedIndex> {
+    let mut indices_set: HashSet<GeneralizedIndex> = HashSet::new();
+    for &index in indices {
+        let helper_indices = get_helper_indices(&[index]);
+        for helper_index in helper_indices {
+            indices_set.insert(helper_index);
+        }
+        let path_indices = get_path_indices(index);
+        for path_index in path_indices {
+            indices_set.remove(&path_index);
+        }
+        indices_set.insert(index);
+    }
+    let mut sorted_indices: Vec<GeneralizedIndex> = indices_set.into_iter().collect();
+    sorted_indices.sort_by_key(|index| format!("{:b}", *index));
+    sorted_indices
+}
+
+pub fn compute_proof_descriptor(indices: &[GeneralizedIndex]) -> Result<Vec<u8>, Error> {
+    let indices = compute_proof_indices(indices);
+    let mut bitstring = String::new();
+    for &index in &indices {
+        let bin_str = format!("{:b}", index);
+        let zeros = bin_str.len() - bin_str.trim_end_matches('0').len();
+        bitstring.push_str(&"0".repeat(zeros));
+        bitstring.push('1');
+    }
+    if bitstring.len() % 8 != 0 {
+        let additional_bits = 8 - (bitstring.len() % 8);
+        bitstring.push_str(&"0".repeat(additional_bits));
+    }
+
+    if let Ok(num) = usize::from_str_radix(&bitstring, 2) {
+        let bytes = num.to_be_bytes();
+        let significant_bytes = (bitstring.len() + 7) / 8;
+        Ok(bytes[bytes.len() - significant_bytes..].to_vec())
+    } else {
+        Err(Error::InvalidDescriptor(bitstring))
+    }
+}
+
+fn compute_bits_from_proof_descriptor(descriptor: &[u8]) -> Result<Vec<bool>, Error> {
+    // Convert bytes to a continuous bit string
+    let bitstring: Vec<bool> = descriptor
+        .iter()
+        .flat_map(|&byte| (0..8).rev().map(move |i| (byte >> i) & 1 == 1))
+        .collect();
+
+    // Find the last '1' in the bitstring
+    let last_one_index = bitstring
+        .iter()
+        .rposition(|&bit| bit)
+        .ok_or(Error::InvalidDescriptor("Descriptor does not contain any '1' bits".to_string()))?;
+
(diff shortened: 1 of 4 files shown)


ralexstokes:
hi @rista404 this is great, thank you!

can you address the linter errors?

otherwise I'd generally lean to merge this soon

ssz-rs/src/merkleization/compact_multiproofs.rs
```diff
@@ -0,0 +1,182 @@
...
+        let mut hasher = Sha256::new();
+        hasher.update(left);
+        hasher.update(right);
+        result.copy_from_slice(&hasher.finalize_reset());
+        Ok(result)
+    }
+}
+
+pub fn verify_compact_merkle_multiproof(
+    nodes: &[Node],
```
ralexstokes, line 130: `nodes` here is the leaf layer from a given merkle tree?

ssz-rs/src/merkleization/compact_multiproofs.rs
```diff
@@ -0,0 +1,182 @@
...
+    descriptor: &[u8],
```
ralexstokes, line 131: what do you think about using a `bitvec` type here instead? then we can skip a lot of `bool` parsing elsewhere in this module

ssz-rs/src/merkleization/compact_multiproofs.rs
```diff
@@ -0,0 +1,182 @@
...
+            ));
+        }
+    }
+
+    Ok(bits)
+}
+
+fn calculate_compact_multi_merkle_root(nodes: &[Node], descriptor: &[u8]) -> Result<Node, Error> {
+    let bits = compute_bits_from_proof_descriptor(descriptor)?;
+    let mut ptr = [0, 0]; // [bit_index, node_index]
```
ralexstokes, line 97: I would probably put these in a `struct` just to encapsulate the details as we traverse the `nodes`

although I'd be fine merging this and adjusting this in a future PR

ssz-rs/src/merkleization/compact_multiproofs.rs
```diff
@@ -0,0 +1,182 @@
...
+    if let Ok(num) = usize::from_str_radix(&bitstring, 2) {
+        let bytes = num.to_be_bytes();
+        let significant_bytes = (bitstring.len() + 7) / 8;
+        Ok(bytes[bytes.len() - significant_bytes..].to_vec())
+    } else {
+        Err(Error::InvalidDescriptor(bitstring))
+    }
+}
+
+fn compute_bits_from_proof_descriptor(descriptor: &[u8]) -> Result<Vec<bool>, Error> {
```
ralexstokes, line 52: could make this much more efficient with just some bit-level arithmetic going byte-by-byte in `descriptor` and can return a `bitvec` type that would be much more compact than `Vec<bool>`

fine to merge this for now and make an issue to refactor

ssz-rs/src/merkleization/multiproofs.rs
```diff
@@ -52,7 +52,7 @@ pub fn calculate_merkle_root(
 ) -> Result<Node, Error> {
     let path_length = get_path_length(index)?;
     if path_length != proof.len() {
-        return Err(Error::InvalidProof)
+        return Err(Error::InvalidProof);
```
ralexstokes, line 55: this repo prefers a nightly lint option that doesn't require trailing `;` here
```suggestion
        return Err(Error::InvalidProof)
```

ssz-rs/src/merkleization/multiproofs.rs
```diff
@@ -89,11 +89,11 @@ pub fn calculate_multi_merkle_root(
     indices: &[GeneralizedIndex],
 ) -> Result<Node, Error> {
     if leaves.len() != indices.len() {
-        return Err(Error::InvalidProof)
+        return Err(Error::InvalidProof);
```
ralexstokes, line 92:
```suggestion
        return Err(Error::InvalidProof)
```

ssz-rs/src/merkleization/multiproofs.rs
```diff
@@ -89,11 +89,11 @@ pub fn calculate_multi_merkle_root(
...
     }
     let helper_indices = get_helper_indices(indices);
     if proof.len() != helper_indices.len() {
-        return Err(Error::InvalidProof)
+        return Err(Error::InvalidProof);
```
ralexstokes, line 96:
```suggestion
        return Err(Error::InvalidProof)
```

ssz-rs/src/merkleization/compact_multiproofs.rs
```diff
@@ -0,0 +1,182 @@
...
+        if bit {
+            count_0_vs_1 -= 1;
+        } else {
+            count_0_vs_1 += 1;
+        }
+
+        // Check mismatch condition at the last index
+        if (count_0_vs_1 < 0) != (i == last_one_index) {
+            return Err(Error::InvalidDescriptor(
+                "Mismatch in count of '0's vs '1's at the last index".to_string(),
```
ralexstokes, line 87: consider making this variant more descriptive

e.g. `InvalidIndexBalanceInDescriptor` that takes no data and then just move this string to the error message

ralexstokes (commented):
generally looks good!

I don't fully get yet how this is different from the other multiproof code, but I think I need to go check the other resources you linked in your first comment


edit: to be clear, I think we can make some of the bit munging more efficient but happy to merge this as-is and refactor later

rista404:
@ralexstokes thanks for the review, I'll go back and make those changes in the same PR! I also highly recommend looking into the spec to get a better understanding of the compact multiproofs.

rista404:
@ralexstokes on a second thought, could we make an issue to refactor the code to use `bitvec`? I would have to see what's the best way to do that as the descriptor is a format defined in the spec, that you can for example send hex encoded over the wire etc.

Still very keen to optimise but I think I would need some more time. If the rest looks okay and you've had the time to look into the spec PR, feel free to merge 🙏🏻 

Thanks again!

rista404:
@ralexstokes ping on this! anything else I can do to help get this merged?

PR #157, opened 2024-07-27
added support for multiproof generation within Prover

### LINKED ISSUE #123: Multiproof generation
Support (multi)proof generation, such that it can be verified with https://github.com/ralexstokes/ssz-rs/blob/43b187ea8b86c79405d587896c764e14ee387f39/ssz-rs/src/merkleization/multiproofs/mod.rs#L138

Something like
```rust
trait Provable {
  fn prove(&self, path: Path<'_>) -> (Witness, Proof)
}
```

where for convenience, `Witness` is the "leaf" data from `self` to prove, and the `proof` is the Merkle multiproof that attests to its validity.

NOTE: there may be overlap with #122

noahfigueras:
I needed multiproof generation to verify multiple fields of data against the beacon block root in the evm. I came with this solution for minimal changes using Prover, not sure if that is acceptable, but works good for me. This is an idea to close #123

### DIFF
--- a/ssz-rs/src/merkleization/proofs.rs
+++ b/ssz-rs/src/merkleization/proofs.rs
@@ -3,8 +3,8 @@ pub use crate::merkleization::generalized_index::log_2;
 use crate::{
     lib::*,
     merkleization::{
-        compute_merkle_tree, GeneralizedIndex, GeneralizedIndexable, MerkleizationError as Error,
-        Node, Path,
+        compute_merkle_tree, multiproofs, GeneralizedIndex, GeneralizedIndexable,
+        MerkleizationError as Error, Node, Path,
     },
 };
 use sha2::{Digest, Sha256};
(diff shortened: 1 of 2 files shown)


ssz-rs/src/merkleization/proofs.rs
```diff
@@ -207,7 +251,7 @@ pub fn is_valid_merkle_branch(
     root: Node,
 ) -> Result<(), Error> {
     if branch.len() != depth {
-        return Err(Error::InvalidProof)
+        return Err(Error::InvalidProof);
```
ralexstokes, line 254:
```suggestion
        return Err(Error::InvalidProof)
```

noahfigueras:
Hi @mempirate! Yeah I agree, the way it's performed it's not really efficient. I'm currently computing a tree to collect each branch for the proof. Ideally, you would want to call prover one time and traverse the tree at once if you keep track of indices and helper indices you can just collect the branches from the leaves. Last time I tried, I had some issues to keep that inside the current Prover implementation, I'll have to rethink this a little.

mempirate:
https://github.com/ralexstokes/ssz-rs/pull/158 also helps with performance by enabling assembly support for hashing. I think combining this with a more efficient multiproof generation algo should already help a lot with proof generation time.

mempirate:
Was able to get proof generation speed down to around 2-3 seconds as a first step, first with `sha2-asm`:
```
Generated proof in 115.117083ms
Verified proof in 10.042µs
```

And with the hashtree_rs crate (https://crates.io/crates/hashtree-rs):
```
Generated multiproof in 2.328004958s
Verified multiproof in 30.042µs
```

PR #158, opened 2024-09-04
`sha2-asm` feature

mempirate:
## Context
This PR implements a feature for enabling the `asm` feature on the `sha2` crate. When testing against a real world payload (generate a proof of transaction inclusion against the transactions root), these are the numbers:

- `sha2-asm` feature disabled:
```
Transactions root: 0x3834068daa909f5553bd06299dc60ccd43fa2c24da3778dda30b6c7fc767f240, num transactions: 129
Index to prove: 26
Generated proof in 855.763416ms
Verified proof in 9.708µs
```
- `sha2-asm` feature enabled:
```
Transactions root: 0x3834068daa909f5553bd06299dc60ccd43fa2c24da3778dda30b6c7fc767f240, num transactions: 129
Index to prove: 26
Generated proof in 133.917375ms
Verified proof in 1.917µs
```

## Conclusion
With `sha2-asm` enabled, proof generation becomes around 6,5 times faster.

### DIFF
--- a/ssz-rs/src/merkleization/proofs.rs
+++ b/ssz-rs/src/merkleization/proofs.rs
@@ -207,7 +207,7 @@ pub fn is_valid_merkle_branch(
     root: Node,
 ) -> Result<(), Error> {
     if branch.len() != depth {
-        return Err(Error::InvalidProof)
+        return Err(Error::InvalidProof);
     }
 
     let mut derived_root = leaf;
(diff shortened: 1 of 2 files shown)


ssz-rs/src/merkleization/proofs.rs
```diff
@@ -207,7 +207,7 @@ pub fn is_valid_merkle_branch(
     root: Node,
 ) -> Result<(), Error> {
     if branch.len() != depth {
-        return Err(Error::InvalidProof)
+        return Err(Error::InvalidProof);
```
ralexstokes, line 210:
```suggestion
        return Err(Error::InvalidProof)
```

ssz-rs/src/merkleization/proofs.rs
```diff
@@ -295,6 +295,34 @@ pub(crate) mod tests {
         assert!(result.is_ok());
     }
 
+    #[test]
```
ralexstokes, line 298: nice! I may refactor this into a Cargo benchmark, but this is fine for now

ralexstokes (approved):
sweet! I thought this was enabled automatically but it looks like this is not supported in this version of this crate

I also like exposing a feature so a user of this library can opt-in or not

ralexstokes:
linter error looks like a new issue from a nightly release -- ill go fix and then we can merge this in

ralexstokes:
@mempirate if you rebase against main ill merge this in

ssz-rs/src/merkleization/proofs.rs
mempirate: Yeah I'm also experimenting with hashtree_rs, could add benchmarks in a follow-up PR!

mempirate:
@ralexstokes this branch should be up to date, nothing to rebase. Maybe I misunderstood

ralexstokes:
hmm yeah im a little confused why the linter is still failing, let me see what I can do -- maybe I just merge anyway as I don't think it will fail on `main`

Issue #160, opened 2024-09-04
Add cargo benchmarks for performance data


ralexstokes:
Can start by converting this test to a benchmark to get perf data for proof gen/verify

https://github.com/ralexstokes/ssz-rs/blob/1df4cd9b849a48c44a6105abc3a38d21cd4fd8d3/ssz-rs/src/merkleization/proofs.rs#L299

EchoAlice:
someone else started working on this, right? If so, i'll close my pr

PR #162, opened 2024-09-06
Create benchmarks for proof generation and verification

EchoAlice:
Addresses issue #160

### DIFF
--- a/ssz-rs/src/lib.rs
+++ b/ssz-rs/src/lib.rs
@@ -82,10 +82,12 @@
 //! ```
 //!
 //! [ssz]: https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md
-#![cfg_attr(not(feature = "std"), no_std)]
 
+#![feature(test)]
+#![cfg_attr(not(feature = "std"), no_std)]
 #[cfg(not(feature = "std"))]
 extern crate alloc;
+extern crate test;
 
 mod array;
 mod bitlist;
--- a/ssz-rs/src/merkleization/proofs.rs
+++ b/ssz-rs/src/merkleization/proofs.rs
@@ -295,8 +295,8 @@ pub(crate) mod tests {
         assert!(result.is_ok());
     }
 
-    #[test]
-    fn test_list_proving() {
+    #[bench]
+    fn bench_proof_generation(b: &mut test::Bencher) {
         let inner: Vec<List<u8, 1073741824>> = vec![
             vec![0u8, 1u8, 2u8].try_into().unwrap(),
             vec![3u8, 4u8, 5u8].try_into().unwrap(),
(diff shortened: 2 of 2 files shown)


EchoAlice:
i'm getting warnings about code unrelated to the PR which is making the build check not pass.  Should i fix these warnings?

ralexstokes:
looks like the issue is directly related to this PR

```bash
error[E0554]: `#![feature]` may not be used on the stable release channel
  --> ssz-rs/src/lib.rs:86:12
   |
86 | #![feature(test)]
   |            ^^^^

For more information about this error, try `rustc --explain E0554`.
```

and it seems like the issue is that the "native" rustc support for benchmarks is not on a stable version of the compiler yet

it looks like a common alternative is `Criterion`, e.g https://github.com/bheisler/criterion.rs?tab=readme-ov-file#quickstart

I think it will also be nicer if we have the benchmarks collected outside of the `src` files, say in a top-level `benches` directory

EchoAlice:
> looks like the issue is directly related to this PR
> 
> ```shell
> error[E0554]: `#![feature]` may not be used on the stable release channel
>   --> ssz-rs/src/lib.rs:86:12
>    |
> 86 | #![feature(test)]
>    |            ^^^^
> 
> For more information about this error, try `rustc --explain E0554`.
> ```
> 
> and it seems like the issue is that the "native" rustc support for benchmarks is not on a stable version of the compiler yet
> 
> it looks like a common alternative is `Criterion`, e.g https://github.com/bheisler/criterion.rs?tab=readme-ov-file#quickstart
> 
> I think it will also be nicer if we have the benchmarks collected outside of the `src` files, say in a top-level `benches` directory

This was what i was referring to:

```
warning: the item `Ordering` is imported redundantly
   --> ssz-rs/src/list.rs:337:17
    |
336 |     use super::*;
    |         -------- the item `Ordering` is already imported here
337 |     use crate::{lib::cmp::Ordering, serialize, U256};
    |                 ^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default

```

ssz-rs/src/lib.rs
```diff
@@ -82,10 +82,12 @@
 //! ```
 //!
 //! [ssz]: https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md
-#![cfg_attr(not(feature = "std"), no_std)]
 
+#![feature(test)]
+#![cfg_attr(not(feature = "std"), no_std)]
 #[cfg(not(feature = "std"))]
 extern crate alloc;
+extern crate test;
```
ralexstokes, line 90: you shouldn't need this anymore

ralexstokes:
> > looks like the issue is directly related to this PR
> > ```shell
> > error[E0554]: `#![feature]` may not be used on the stable release channel
> >   --> ssz-rs/src/lib.rs:86:12
> >    |
> > 86 | #![feature(test)]
> >    |            ^^^^
> > 
> > For more information about this error, try `rustc --explain E0554`.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and it seems like the issue is that the "native" rustc support for benchmarks is not on a stable version of the compiler yet
> > it looks like a common alternative is `Criterion`, e.g https://github.com/bheisler/criterion.rs?tab=readme-ov-file#quickstart
> > I think it will also be nicer if we have the benchmarks collected outside of the `src` files, say in a top-level `benches` directory
> 
> This was what i was referring to:
> 
> ```
> warning: the item `Ordering` is imported redundantly
>    --> ssz-rs/src/list.rs:337:17
>     |
> 336 |     use super::*;
>     |         -------- the item `Ordering` is already imported here
> 337 |     use crate::{lib::cmp::Ordering, serialize, U256};
>     |                 ^^^^^^^^^^^^^^^^^^
>     |
>     = note: `#[warn(unused_imports)]` on by default
> ```

I see a different error, need to update the MSRV

PR #170, opened 2024-12-06
perf: use precomputed zero hashes in process_subtree

estensen:
Stacked on https://github.com/ralexstokes/ssz-rs/pull/169

A lot of leaves are empty and we don't need to hash them or their parents.

Precomputing zero hashes for leafs
```
Prove Benchmark - File: benches/21315748.json - size 247/123
                        time:   [96.891 ms 97.480 ms 98.564 ms]
                        change: [-36.824% -36.084% -35.371%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking Prove Benchmark - File: benches/21327802.json - size 261/130: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 5.3s or enable flat sampling.
Prove Benchmark - File: benches/21327802.json - size 261/130
                        time:   [96.672 ms 97.006 ms 97.291 ms]
                        change: [-38.752% -37.874% -37.086%] (p = 0.00 < 0.05)
                        Performance has improved.
```

On the layer above we can do a similar thing; if both the leaves are zero hashes we can precompute their parent as well.

Precomputing hash for parent of zero leafs
```
Prove Benchmark - File: benches/21315748.json - size 247/123
                        time:   [63.227 ms 63.584 ms 63.845 ms]
                        change: [-36.239% -35.537% -34.850%] (p = 0.00 < 0.05)
                        Performance has improved.

Prove Benchmark - File: benches/21327802.json - size 261/130
                        time:   [64.678 ms 66.169 ms 66.956 ms]
                        change: [-33.989% -33.024% -31.938%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Precomputing hash for grandparent of zero leafs
```
Prove Benchmark - File: benches/21315748.json - size 247/123
                        time:   [48.241 ms 48.500 ms 48.906 ms]
                        change: [-24.267% -23.604% -22.946%] (p = 0.00 < 0.05)
                        Performance has improved.

Prove Benchmark - File: benches/21327802.json - size 261/130
                        time:   [48.734 ms 48.919 ms 49.099 ms]
                        change: [-25.885% -24.824% -23.838%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Precomputing hash for grandgrandparent of zero leafs
```
Prove Benchmark - File: benches/21315748.json - size 247/123
                        time:   [43.100 ms 43.391 ms 43.802 ms]
[...]

PR #169, opened 2024-12-02
perf: compute subtrees in parallel when computing proof

estensen:
Merkle trees are binary trees, enabling parallel processing of subtrees. By parallelizing 8 subtrees, we get ~83% reduction in processing time for two mainnet blocks. This implementation detects the number of available CPU cores and runs either serial processing, parallel 4 subtrees (RPi4) or 8 subtrees.

Process 2 subtrees in parallel
```
Prove Benchmark - File: benches/21315748.json - size 247/123
                        time:   [250.73 ms 251.78 ms 253.11 ms]
                        change: [-44.361% -43.919% -43.531%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe

Prove Benchmark - File: benches/21327802.json - size 261/130
                        time:   [251.22 ms 252.05 ms 252.89 ms]
                        change: [-44.012% -43.797% -43.578%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Process 4 subtrees in parallel compared to 2
```
Prove Benchmark - File: benches/21315748.json - size 247/123
                        time:   [158.15 ms 159.54 ms 161.30 ms]
                        change: [-42.477% -41.625% -40.843%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking Prove Benchmark - File: benches/21327802.json - size 261/130: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 8.7s or enable flat sampling.
Benchmarking Prove Benchmark - File: benches/21327802.json - size 261/130: Collecting 10 samples in estima
Prove Benchmark - File: benches/21327802.json - size 261/130
                        time:   [158.71 ms 159.15 ms 159.40 ms]
                        change: [-43.690% -43.042% -42.435%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 10 measurements (30.00%)
  2 (20.00%) low mild
  1 (10.00%) high severe
```

Process 8 subtrees in parallel compared to 4
```
Prove Benchmark - File: benches/21315748.json - size 247/123
                        time:   [92.659 ms 93.286 ms 93.841 ms]
                        change: [-41.520% -40.689% -39.987%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking Prove Benchmark - File: benches/21327802.json - size 261/130: Warming up for 3.0000 s
[...]

PR #171, opened 2024-12-06
feat: implement multiproof generation

### LINKED ISSUE #123: Multiproof generation

estensen:
Resolves https://github.com/ralexstokes/ssz-rs/issues/123
An alternative implementation to https://github.com/ralexstokes/ssz-rs/pull/157 to only build the tree once when generating multi proofs.

Would it be more ergonomic to change the API and decoupling tree building from proof generation to make implementation more modular and easier to benchmark? The way the code is currently structured if you wanna create a proof and then a multiproof you need to build the tree twice. Alternatively the tree itself could be cached.

Merkleproofs had no tests, so I added some basic ones to make sure I don't break it too bad while optimizing.

Most of the time is spent on building the tree, not generating the proofs themselves, so stacking https://github.com/ralexstokes/ssz-rs/pull/169 and https://github.com/ralexstokes/ssz-rs/pull/170 should get down to less than 30ms:
```
MultiProve Benchmark - File: benches/21315748.json - size 247/Generate Single Proof
                        time:   [446.69 ms 447.43 ms 448.28 ms]
MultiProve Benchmark - File: benches/21315748.json - size 247/Verify Single Proof
                        time:   [6.8539 µs 6.8755 µs 6.8968 µs]

MultiProve Benchmark - File: benches/21315748.json - size 247/Generate Ten Proofs
                        time:   [453.34 ms 454.12 ms 455.01 ms]
MultiProve Benchmark - File: benches/21315748.json - size 247/Verify Ten Proofs
                        time:   [26.836 µs 27.267 µs 28.143 µs]

MultiProve Benchmark - File: benches/21327802.json - size 261/Generate Single Proof
                        time:   [459.21 ms 461.32 ms 464.11 ms]
MultiProve Benchmark - File: benches/21327802.json - size 261/Verify Single Proof
                        time:   [7.0034 µs 7.0473 µs 7.1087 µs]

MultiProve Benchmark - File: benches/21327802.json - size 261/Generate Ten Proofs
                        time:   [462.67 ms 465.52 ms 468.85 ms]
MultiProve Benchmark - File: benches/21327802.json - size 261/Verify Ten Proofs
                        time:   [27.273 µs 27.342 µs 27.452 µs]
```


ssz-rs/src/merkleization/mod.rs
```diff
@@ -1,5 +1,5 @@
 pub mod generalized_index;
-mod merkleize;
+pub mod merkleize;
```
estensen: prob revert

